PR #3, opened 2019-05-21
code refactor

yashnalla:
Changed Readme
Pulled Utility functions to own module


akamai/update_api.py
```diff
@@ -145,30 +125,28 @@ def activateVersion(version_number, env="STAGING"):
...
 
-def akamaiGet(url):
-    return s.get(urljoin(baseurl, url)).content
-
-def akamaiPost(url, body):
-    return s.post(urljoin(baseurl, url), json=body).content
 
-def akamaiPut(url, body):
-    return s.put(urljoin(baseurl, url), json=body).content
+def main():
```
kruai: Having the script in a `main` method makes the code look so much cleaner :smile:

akamai/update_api.py
```diff
@@ -145,30 +125,28 @@ def activateVersion(version_number, env="STAGING"):
...
-    return s.get(urljoin(baseurl, url)).content
-
-def akamaiPost(url, body):
-    return s.post(urljoin(baseurl, url), json=body).content
 
-def akamaiPut(url, body):
-    return s.put(urljoin(baseurl, url), json=body).content
+def main():    
+    # Set up connectivity
+    s = requests.Session()
```
kruai: Ah, now I see where `s` went... it does make things cleaner to have it in the `main` method, but I made it global before since I was using it in a static fashion. Changing the structure in this way means that you'd have to pass `s` as an argument all the way down to the function where it's actually used.

Personally, I think I prefer it as a global static variable.

yashnalla: Sure. I just realized that I never tested this after I shifted S to the main function. It was the very last thing I did but probably should've paid more attention

PR #4, opened 2019-05-23
Automatic Akamai config

kruai:
Automatically generates an Akamai config based on the source of truth `main.yml`.


kruai:
Don't worry about the number of commits - there was a period where I had to experiment extensively with Jenkins :stuck_out_tongue:

main.yml
```diff
@@ -46,6 +47,11 @@ insights:
   deployment_repo: https://github.com/RedHatInsights/insights-advisor-frontend-build
   frontend_paths:
   - /insights
+  frontend_exclude:
```
iphands: Why do we need this?
I noticed you adding this to the manual configs, but I dont think its really necessary.
Shouldnt the "last match wins" in Akamai prevent the need for this?

kruai: It's "last match wins", but multiple matches can occur, which has caused us issues before. That's why we changed our rules for insights to match `/insights/*` but exclude `/insights/inventory/*`, `/insights/remediations/*`, etc.
We originally decided to do it this way in order to be extra explicit with our app rules; we might be able to avoid using exclusions by forcing the rules to be created in a certain order, though it would be less explicit.

kruai: Come to think of it, I might be able to generate these exclusions just based on the paths of the other apps in the YAML. If you want, I can take a stab at that once I'm done with my other work today.

iphands:
> Come to think of it, I might be able to generate these exclusions just based on the paths of the other apps in the YAML. If you want, I can take a stab at that once I'm done with my other work today.

Oh man that sounds awesome.
In spandex we dealt with order buy sorting via string length...

I feel like if we sort of be a string length then lay out the rules in that order we might be fine.

But test it all via the smoke tester

kruai:
UNLIMITED POWERRRRR. It now dynamically creates rule trees based on the data in `data/environments.yml`, and determines the frontend paths it should exclude based on the paths of the other apps within the configuration :smile:

PR #5, opened 2019-06-03
Enhancements/chrome nav

kruai:
I changed `main.yml` into a list, and provided detail that I'm using to generate the navigation in Chrome. I tidied it up some, but I would love to clean it up more if possible.

### DIFF
--- a/akamai/update_api.py
+++ b/akamai/update_api.py
@@ -150,7 +150,7 @@ def generateExclusions(frontend_path, config):
     return exclusions
 
 def generateConfigForBranch(branch):
-    config = util.getYMLFromUrl("https://raw.githubusercontent.com/RedHatInsights/cloud-services-config/enhancements/chrome-nav/main.yml".format(branch))
+    config = util.getYMLFromUrl("https://raw.githubusercontent.com/RedHatInsights/cloud-services-config/{}/main.yml".format(branch))
     # For every app in config, check all other apps to see if they have a frontend_path that contains its frontend_paths.
     for key in (x for x in config.keys() if "frontend" in config[x] and "paths" in config[x]["frontend"]):
         exclusions = []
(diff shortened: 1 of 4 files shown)


akamai/update_api.py
```diff
@@ -142,20 +143,20 @@ def activateVersion(version_number, env="STAGING"):
 
 def generateExclusions(frontend_path, config):
     exclusions = []
-    for app in (x for x in config if "frontend_paths" in config[x] and frontend_path not in config[x]["frontend_paths"]):
-        for path in config[app]["frontend_paths"]:
+    for key in (x for x in config.keys() if "frontend" in config[x] and "paths" in config[x]["frontend"] and frontend_path not in config[x]["frontend"]["paths"]):
```
kruai: I want to use `for key, app in ()` but I'm not sure how to do that with the filtering I'm doing in the generator expression. Anyone know how to do that? I think it would look cleaner.

iphands: I would consider pulling the python https://pypi.org/project/pylodash/ dep

main.yml
```diff
@@ -1,98 +1,277 @@
...
+      - id: drift
+      - id: inventory
+      - id: remediations
+  top_level: true
+
+settings-hybrid:
+  name: Settings
+  frontend:
+    sub_apps:
+      - id: catalog-sources
```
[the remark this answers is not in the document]
kruai: Most likely!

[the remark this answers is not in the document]
kruai: Answer: no, it's not different. Seems like we want different titles in different places, though. I'll make an update

akamai/update_api.py
kruai: I did pull it in for a while to see if it would help. I couldn't find a way to use these functions to alleviate the issue, though. If you have an idea, can you please let me know?

PR #7, opened 2019-06-18
Add migration-analytics to main.yml

carlosthe19916:
This information should be enough to configure migration-analytics


kruai:
These changes don't line up with the structure we require in our YAML file. Please make a new branch off of `master`, follow the instructions [here](https://github.com/RedHatInsights/cloud-services-config#adding-config-for-new-apps), and open another PR.

PR #8, opened 2019-06-18
Explain the configuration in the README

kruai:
In order to help people make PRs for new and updated apps, I wanted to explain the source of truth's configuration in detail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,105 @@
 
 This repo deals with the high-level configuration of Cloud Services. `main.yml` contains the source of truth for CS apps, and the `akamai` folder deals with updating our Akamai configuration.
 
+# Adding Config for New Apps
+
+To enable a new app in our environments, you need to create configuration for it in `main.yml`, and then create a PR to merge it in. 
+Here is some example configuration that demonstrates the structure, using all required and optional properties:
+```
+{app_id}: 
+    title: App Title
+    api:
+        versions:
+            - v1
+            - v2
+    channel: '#some-slack-channel'
+    description: App Title is a cool app that does business things for its users.
+    deployment_repo: https://github.com/app-deployment-repo-url
+    disabled_on_stable: true
+    docs: https://link.to.docs.com/docs
+    frontend:
+        title: App Title Override
+        paths:
+            - /example/path
+            - /another/example/path
+        reload: reload/path
+        sub_apps:
+            -   id: app_id_1
+                title: Some Sub App
+                default: true
+            -   id: app_id_2
+                title: Another Sub App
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -2,14 +2,138 @@
...
+For example, let's say your app `ex_app_id` is a sub-app of `parent_app`, which has a frontend path of `/parent-app`. If you add a frontend path of `/ex-app`, your app will be available at `/parent-app/ex-app`.
+
+#### app_id.frontend.sub_apps
+
+If your app is a parent to any other apps, those apps should be listed here. 
+Also, if your app has a parent app, or is listed under one of the top-level bundles (e.g. Insights, RHEL, Hybrid), you should add your app to the appropriate sub_apps list.
+
+#### app_id.frontend.reload
+
+This property is almost never needed, and will likely be deprecated. This property is commonly used for Settings apps, and tells Chrome where to navigate on reload.
```
karelhala: I am not sure if we are going to deprecate it, this prop should be documented as a way to change URL where you app is located. So for instance app is located under some first level and is not managed by this first level app you should set it up. Currently only for settings because settings is not really a application and the URL will look like `/settings/something` or catalog has rbac that is specific application located under `catalog` app. I would phrase this as specific more direct URL if you want to redirect automatic generation of URL.

README.md
```diff
@@ -2,14 +2,138 @@
...
+
+#### app_id.frontend.suppress_id
+
+This property is almost never needed, and will likely be deprecated. This property is used if the app isn't a real app on disk, and only exists for navigation purposes.
```
karelhala: Same as above I don't really think we can depracate this because we will probably want some apps to be groupped under one first level nav, but this nav has no app to it (like settings).

README.md
kruai: Yeah - I know we need it right now, but when I was writing it, I wondered if there was better way to determine this information. I should definitely reword this for now.

README.md
kruai: Yeah - we can't deprecate this now because of the way we're coding it. I feel like there's a way we could get by without making users define this property. For instance, we could probably infer `suppress_id = True` if `deployment_repo` isn't defined for an app.

I don't want to say this definitively until the Chrome nav auto-generation is done, but I still think it's likely we'll get rid of this property at some point

kruai: I'm removing the part about it rarely being needed and us deprecating it. We can keep that info internal :)

README.md
kruai: I'm removing the first sentence for this one, too. Users don't need to know that right now.

PR #9, opened 2019-06-19
Cache HTTP errors to reduce load spikes; rule name update

kruai:
I talked to Kyle Lape and found that when Akamai has lag spikes, our site could hang for upwards of 20 seconds just to finally present the user with a 404. Akamai reps recommended caching HTTP errors, so I'm doing that (for frontend only, not API).

I also updated one of the rule labels, just because we're not pre-release anymore.

### DIFF
--- a/akamai/data/base_rules.json
+++ b/akamai/data/base_rules.json
@@ -209,11 +209,19 @@
                         "options": {
                             "behavior": "ALWAYS"
                         }
+                    },
+                    {
+                        "name": "cacheError",
+                        "options": {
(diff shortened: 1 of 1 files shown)


kruai:
I also added config for the cert that Akamai user Shwetha Bangera added to our config. I'm not sure what it's for yet, but after talking to Akamai over chat, they've opened Akamai case #F-CS-3517108 to get us info on that. I think it's safe to assume they added the cert for a good reason, and we should keep it until we determine otherwise.

PR #10, opened 2019-06-20
Add migration-analytics configuration

carlosthe19916:
Migration Analytics will have his own bundle, not accessible directly from landing page. In the future, we could create a button on the landing page that would bring users to this comparison page.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -288,4 +288,5 @@ migration-analytics:
     paths:
       - /migration-analytics
   git_repo: https://github.com/project-xavier/xavier-ui
+  mailing_list: migration-analytics-devel@redhat.com
   top_level: true
--- a/main.yml
+++ b/main.yml
@@ -286,7 +286,7 @@ migration-analytics:
   disabled_on_stable: true
   frontend:
     paths:
-      - /migration-analytics
+      - /apps/migration-analytics
   git_repo: https://github.com/project-xavier/xavier-ui
   mailing_list: migration-analytics-devel@redhat.com
   top_level: true
--- a/main.yml
+++ b/main.yml
@@ -275,3 +275,18 @@ vulnerability:
     title: Vulnerability
     paths:
       - /rhel/vulnerability
+
+migration-analytics:
+  title: Migration Analytics
+  api:
+    versions:
+      - v1
+  deployment_repo: https://github.com/RedHatInsights/xavier-ui-deploy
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -275,3 +275,18 @@ vulnerability:
...
+  title: Migration Analytics
+  api:
+    versions:
+      - v1
+  deployment_repo: https://github.com/RedHatInsights/xavier-ui-deploy
+  disabled_on_prod: true
+  disabled_on_stable: true
+  frontend:
+    paths:
+      - /migration-analytics
```
karelhala: If you want to be located under `/apps/migration-analytics` you should add full path here.

carlosthe19916: @karelhala Thanks for reviewing this PR. I just want to double check your suggestion.
Do you want me to change `/migration-analytics` for `/apps/migration-analytics`? I'm asking because https://github.com/RedHatInsights/cloud-services-config/blob/master/main.yml does not contain any application which uses `/apps` as a prefix.

karelhala: Yes that is correc. We don't have any application accessible trough `/apps` prefix because all of them are in some bundle, however your app is specific that you don't have any bundle.

Without `/apps` the URL would look like `https://cloud.redhat.com/migration-analytics`, however if you include it the outcome would be `https://cloud.redhat.com/apps/migration-analytics`.

There is also a PR ready to review in chrome part to allow showing these applications (application without any bundle) in specific way https://github.com/RedHatInsights/insights-chrome/pull/324

carlosthe19916: @karelhala Thank you very much; now I understand. I've already changed the code and updated the PR. I hope it works now.

kruai: @karelhala Since it's its own bundle, the usual pattern would make me expect this app to be deployed to `cloud.redhat.com/migration-analytics`. Every app has their assets deployed to `/apps/<appname>`, so you can always access the app through that link.

In my opinion, if you're ok with the deployment URL being `cloud.redhat.com/migration-analytics`, we should leave the frontend path the way it was. They'll still be able to access their app through `/apps/migration-analytics` if they need to.

karelhala: Yeah, but that is not accessible on CI :disappointed:

Let's go with `/apps` (will be much better experience once https://github.com/RedHatInsights/insights-chrome/pull/324 is merged) we can change it at any time in future though.

PR #13, opened 2019-07-02
Add ruledev bundle and app

karelhala:
~~I am not sure about the id of sub app. The url should look like `cloud.redhat.com/ruledev` that's why the ID is empty.~~ Fixed

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -233,15 +233,8 @@ ruledev:
   channel: '#flip-mode-squad'
   deployment_repo: https://github.com/RedHatInsights/ruledev-frontend-build
   frontend:
-    title: Rule Development
     paths:
-      - /ruledev
-    sub_apps:
-      - id: 'ruledev'
-        title: Home
-        default: true
-        group: ruledev
-  top_level: true
+      - /staging/ruledev
 
 settings-hybrid:
   title: Settings
@@ -292,6 +285,7 @@ staging:
   frontend:
     sub_apps:
       - id: migration-analytics
+      - id: ruledev
   top_level: true
 
 tower-analytics:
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -228,6 +228,21 @@ rhel:
...
   top_level: true
 
+ruledev:
+  title: Rule Development
+  channel: '#flip-mode-squad'
+  deployment_repo: https://github.com/RedHatInsights/ruledev-frontend-build
+  frontend:
+    title: Rule Development
+    paths:
+      - /ruledev
```
kruai: This should be the full path to the app. If we're putting it in the Staging bundle, that's `/staging/ruledev`.

main.yml
```diff
@@ -228,6 +228,21 @@ rhel:
...
+    sub_apps:
```
kruai: Is this an app group? This field should only exist if you want a sub app that exists under the `ruledev` app.

main.yml
```diff
@@ -228,6 +228,21 @@ rhel:
...
+      - id: 'ruledev'
```
kruai: If you add a sub app with an ID that points to the original app, we'll end up with an infinite loop when trying to build the navigation.

main.yml
```diff
@@ -228,6 +228,21 @@ rhel:
...
+        title: Home
+        default: true
+        group: ruledev
+  top_level: true
```
kruai: `top_level` should only be defined if this is a bundle, or something at that level. I'm against that for this app, unless we hear from the higher-ups that `ruledev` needs to be its own separate category.

karelhala:
@kruai not for now at least, we are going to query JIRA directly (at least for now)

PR #19, opened 2019-07-22
Update docs to full name documentation allow accessing /docs to api

karelhala:
* Rename api docs to api documentation
* Rename cloud Services documentation to documentation
* When user hits `/docs` redirect him to `/docs/api`

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -9,7 +9,7 @@ approval:
       - /hybrid/catalog/approval
     reload: catalog/approval
 
-api:
+api-docs:
   title: API documentation
   deployment_repo: https://github.com/RedHatInsights/api-frontend-build
   frontend:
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -10,11 +10,12 @@ approval:
     reload: catalog/approval
 
 api:
```
kruai: I know I approved this ID earlier, but I've been thinking on it. `api` is really vague and is confusing to see in this list of apps. Can we change the ID to something like `api-docs` instead?

main.yml
```diff
@@ -96,7 +97,7 @@ dashboard:
       - /rhel/dashboard
 
 docs:
```
kruai: Since `docs` is really vague too, I would love to update this to something like `cs-docs`. I'm open to other naming suggestions too. Any ideas?

karelhala: Good point!

main.yml
karelhala: Yeah the api felt a bit fishy for me as well. I'll change it to api docs, good point.

PR #24, opened 2019-08-01
Activate Subscription Reporting

cdcabrera:
## Description
Subscription Reporting bundle, remove `disabled`.

## Notes
- remove `disabled_` config settings. 
- create as top level without a card on the landing page, should be accessible from direct link only. may need some input on how to structure this. it is a config so feels pretty open... 

   Since it appears the config aims to have landing areas standalone'ish, an alternative possibility we came up with is here. We currently only have one path/route:

   ```
   subscriptions:
     title: Subscription Reporting
     channel: '#subscriptions'
     frontend:
       sub_apps:
         - id: subscription-rhel
           default: true
     top_level: true
   
   subscriptions-rhel:
     title: Red Hat Enterprise Linux
     api:
       versions:
         - v1
     frontend:
       paths:
         - /subscriptions/rhel
   ```

@kruai @ryelo @karelhala

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -320,19 +320,18 @@ staging:
       - id: subscription-reporting
   top_level: true
 
-subscription-reporting:
+subscriptions:
   title: Subscription Reporting
   api:
     versions:
       - v1
   channel: '#subscriptions'
   deployment_repo: https://github.com/RedHatInsights/curiosity-frontend-build
-  disabled_on_prod: true
-  disabled_on_stable: true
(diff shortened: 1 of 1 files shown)


karelhala:
The second one is more syled based on how we style other stuff, so I guess that might be a bit better to use.

cdcabrera:
After catching up with @kruai we're **going to delay this update** until the structural changes for the repo take place.

PR #25, opened 2019-08-01
Insights will set its own active app (so we only highlight it once)

AllenBW:
~~Cant merge this till https://github.com/RedHatInsights/insights-advisor-frontend/pull/385 makes it in~~ Its in, this is g2g.


AllenBW:
https://github.com/RedHatInsights/insights-advisor-frontend/pull/385 was merged so this one is g2g!

PR #27, opened 2019-08-05
WIP Update Subscription Reporting

cdcabrera:
## Description
Subscription Reporting bundle. Once we get this format worked out we'll open a PR against `ci-stable` per the README

## Notes
- create as top level **without** a card on the landing page, should be accessible from direct link only.

    Screen Shot 2019-08-05 at 11 37 30 AM 

   Since it looks like the config aims to have landing areas we have an alternative config layout below. A little unsure of one thing since we currently only have the one route `[app]/rhel`... 
   - Updating App ID appears obvious, we'd just update towards `subscriptions-rhel`
   - However, since we're splitting the app out using the below config were a little unsure what path our app would be hosted at. Need confirmation, would it still be `/apps/subscriptions`?
   
   #### Alternative
   ```
   subscriptions:
     title: Subscription Reporting
     channel: '#subscriptions'
     frontend:
       sub_apps:
         - id: subscription-rhel
           default: true
     top_level: true
   
   subscriptions-rhel:
     title: Red Hat Enterprise Linux
     api:
       versions:
         - v1
     frontend:
       paths:
         - /subscriptions/rhel
   ```

@kruai @ryelo @karelhala

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -313,10 +313,9 @@ staging:
     sub_apps:
       - id: migration-analytics
       - id: ruledev
-      - id: subscription-reporting
   top_level: true
 
-subscription-reporting:
+subscriptions:
   title: Subscription Reporting
   api:
     versions:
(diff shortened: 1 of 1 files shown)


kruai:
We're avoiding marking apps as "top level" bundles to avoid clutter, but it sounds like that's the explicit goal here. @iphands do you already know about this?

cdcabrera:
So just throwing this out there @kruai since we're not going to have a card, landing area, and it's supposed to be a accessible through a direct link only. I could just remove the `top level` config option this first round? That would help kick the decision down the road a little...

cdcabrera:
Discussion, closing as of 20190805

PR #31, opened 2019-08-16
added migrations bundle

ryandillinfelton:
Adding entry for the new migrations bundle in the **SOURCE OF TRUTH** Opening PR early for easier troubleshooting.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -161,6 +161,20 @@ landing:
   channel: '#flip-mode-squad'
   deployment_repo: https://github.com/RedHatInsights/landing-page-frontend-build
 
+migrations:
+  title: Migrations
+  api: 
+    versions:
+      - v1
+  frontend: 
+    title: Migrations
+    sub_apps:
+      - id: migration-analytics
+        title: Migration Analytics
+        group: migrations
+        default: true
+  top_level: true
+
 migration-analytics:
   title: Migration Analytics
   api:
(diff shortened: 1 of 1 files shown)


ryelo (changes requested):
https://github.com/RedHatInsights/cloud-services-config/blob/e1f567c18816b852b3122b382d1314bd591381e4/main.yml#L174 needs to be updated

main.yml
```diff
@@ -175,6 +175,21 @@ migration-analytics:
...
   mailing_list: migration-analytics-devel@redhat.com
 
+migrations:
+  title: Migration Services
+  api:
+    versions:
+      - v1
+  deployment_repo: https://github.com/RedHatInsights/xavier-ui-deploy
+  frontend:
+    title: Migration
```
ryelo: Should this be `Migrations`?

main.yml
```diff
@@ -161,6 +161,20 @@ landing:
...
+migrations:
+  title: Migrations
+  api: 
+    versions:
+      - v1
+  frontend: 
+    title: Migrations
+    sub_apps:
+      - id: migration-analytics
+        title: Migration Analytics
```
kruai: Supplying a `title` is only necessary if the sub-app isn't defined elsewhere in the config; this field is for things like `/insights/overview`, which aren't a separate app and just act as routing for `/insights`.

main.yml
```diff
@@ -161,6 +161,20 @@ landing:
...
+        group: migrations
```
kruai: I may be mistaken, but I don't think you need the `group` tag for this one.

main.yml
```diff
@@ -171,7 +185,8 @@ migration-analytics:
   disabled_on_stable: true
   frontend:
     paths:
-      - /staging/migration-analytics
```
kruai, line 202: If you're removing it from `/staging`, you should also remove the reference from the Staging bundle.

kruai (changes requested):

main.yml
ryandillinfelton: removed

main.yml
ryandillinfelton: Removed

main.yml
ryandillinfelton: Removed reference from staging sub_apps

kruai (approved):
Looks good now!

ryelo (approved):

PR #32, opened 2019-08-16
update to left-hand nav for system comparison

beav:
This commit renamed "System Comparison" to "Drift Analysis". It also
creates two submenu items, one for "System Comparison", and one for
"Baselines".


karelhala (approved):
Looking good!

PR #33, opened 2019-08-20
Add tags and isBeta to some APIs

karelhala:
@kruai I've rather added `isBeta` to all env that are supposed to be available on beta, if you want I can remove it from some places (like migrations, ansible tower, subscriptions, ruledev and hooks).

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -62,6 +62,7 @@ cost-management:
   api:
     versions:
       - v1
+    isBeta: true
   channel: '#koku'
   deployment_repo: https://github.com/RedHatInsights/cost-management-build
   description: >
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -194,9 +198,6 @@ landing:
 
 migrations:
   title: Migrations
-  api:
```
karelhala, line 197: @ryandillinfelton Migrations is top level nav with children right? Because I don't think that it has api (migration-analytics on other hand has xavier api).

kruai: Good catch. Bundles shouldn't have an associated API afaik

kruai (approved):

PR #34, opened 2019-08-21
Add new app topological inventory to hybrid section

karelhala:


main.yml
```diff
@@ -331,6 +332,17 @@ subscriptions:
       - /staging/subscriptions
   git_repo: https://github.com/RedHatInsights/curiosity-frontend
 
+topological-inventory:
+  title: Topological inventory
```
kruai: Inventory should probably be capitalized

main.yml
```diff
@@ -331,6 +332,17 @@ subscriptions:
...
+  api:
+    versions:
+      - v1.0
```
kruai, line 339: All other APIs are just `v1`. Is this one really `v1.0`?

kruai (changes requested):

main.yml
karelhala: Shouldn't all things in PF4 sentence case?

main.yml
karelhala: Yup, it is `v1.0` the API docs allow both.

main.yml
kruai: I don't know what you mean - what does PF4 have to do with this? All other apps have their titles in title case. If they're specifically telling you they want it to be sentence case, I can approve it, but it's going to look extremely weird in the sidenav when everything else is capitalized.

ryelo: No. Titles are capitalized since they're a proper app name. If that has changed, then all of them should be updated. My new mockups with landing page stuff has titles in caps.

karelhala: Well PF4 defined that everything should be sentence case, but yeah other apps has all upper case, I'll update it.

kruai: Cool. If we decide to start following those guidelines, we can make a change for all of them at once.

kruai (approved):
Looks good!

main.yml
ryelo: @karelhala Not "everything". As per their guidelines:
```
Sentence style

Use sentence-style capitalization for all text elements, except:

    Product names and proper nouns
```

ryelo (approved):

ryelo:
Just so this doesn't get buried in the outdated comment and for future reference:

https://www.patternfly.org/v4/design-guidelines/content/grammar-and-terminology#capitalization
```
Sentence style

Use sentence-style capitalization for all text elements, except:

    Product names and proper nouns
```

All of the apps here should *not* use sentence style/case. Each word should be capitalized.

PR #36, opened 2019-08-26
Adds ansible bundle

tahmidefaz:


tahmidefaz:
Looks like they don't have a deploy repo setup already. Hence, there's no mention of it. Let me know if I need to change anything!

main.yml
```diff
@@ -1,3 +1,12 @@
+ansible:
+  title: Ansible
+  frontend:
+    title: Ansible
```
kruai: `frontend.title` is an override; you only need it if you need the UI to display the name differently than the main `title` field.

kruai (changes requested):
You might only need the title change, but I'm not going to approve just yet; I'm debugging an issue with Ryan F and the config in your PR looks very similar. Once I figure out what's happening with `/migrations`, I'll approve or request more changes

ryandillinfelton (commented):
Looks fine to me. I'll come take a look again after we figure out what's going on with the migrations bundle so we can make sure it won't happen with this too

kruai (approved):
Now that we know migration-analytics is actually correct, I think this one is going to work fine.

ryelo (approved):

ryandillinfelton (approved):
LGTM

kruai:
Your wish is my command

PR #39, opened 2019-08-29
Add tower analytics to the ansible bundle

ryelo:
and updating the name to remove "Ansible"


kruai (approved):
Simple. Also I'm surprised tower-analytics wasn't already in a bundle, like staging!

ryandillinfelton (approved):

PR #45, opened 2019-09-05
OpenShift: Update default sub_app, update title

elad661:
We're using the "OCM" name everywhere now, and we want to use / instead of /clusters

cc: @karelhala @cben


karelhala (approved):

jhernand:
@kruai We would like to push this to production as soon as possible. What do we need to do?

PR #51, opened 2019-09-11
Merge beta -> stable

kruai:
Created a `ci-beta-rc` branch to resolve any conflicts before merging to `ci-stable`.
Some apps that were previously beta-only are included in this PR; PLEASE let me know if we don't want to include any of these in Stable yet.


kruai:
Apps added:
- ansible (bundle)
- automation-analytics
- automation-hub
- hooks
- migrations (bundle)
- migration-analytics
- settings-ansible

I'm mostly concerned about `hooks`. Do we want to add that into Stable yet?

tahmidefaz (approved):

main.yml
```diff
@@ -182,7 +275,7 @@ openshift:
     paths:
       - /openshift
     sub_apps:
-      - id: clusters
```
ryandillinfelton, line 185: Does this not need an ID?

kruai: They wanted to configure it so that there was no path parameter for Clusters (so it's just /openshift). Since Clusters isn't an app defined elsewhere in the config, it's not technically required.

ryandillinfelton (approved):

main.yml
ryandillinfelton: Okay, that makes sense!

You may want to change overview on [line 210](https://github.com/RedHatInsights/cloud-services-config/blob/9a90fafd043dcd71f22781570a626513e612721e/main.yml#L210) to make it consistent with [line 128](https://github.com/RedHatInsights/cloud-services-config/blob/9a90fafd043dcd71f22781570a626513e612721e/main.yml#L128) since it also isn't defined on its own, but I don't think it's critical or anything.

kruai: For Insights, though, Overview appends a path parameter to the end. So while Insights' Overview is at `/insights/overview`, Cost Management's Overview is at `/hybrid/cost-management` (no ID appended).

ryelo (approved):

PR #60, opened 2019-09-23
Reload approval app

rvsia:
**Description**

Approval is not a part of Catalog, so the `reload` option is added.

@karelhala

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -28,6 +28,7 @@ approval:
     paths:
       - /hybrid/catalog/approval
       - /ansible/catalog/approval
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -80,6 +81,7 @@ catalog:
       - id: orders
         title: Orders
       - id: approval
+        reload: catalog/approval
```
karelhala: No need for another reload, it is enough in the `approval` item

karelhala (approved):

PR #72, opened 2019-09-27
remove migrations in ci-stable

ryelo:
we missed this one

PR #73, opened 2019-09-27
Remove migrations from stable

karelhala:


karelhala:
Closing because @ryelo was faster with https://github.com/RedHatInsights/cloud-services-config/pull/72

PR #79, opened 2019-10-11
Added products to catalog sub menu.

Hyperkid123:
We are adding new sub menu in catalog called Products.

cc @gmcculloug


Hyperkid123:
@karelhala can you take a look?

karelhala (approved):

PR #85, opened 2019-10-23
add dev setup instructions to readme

beav:


README.md
```diff
@@ -132,3 +132,49 @@ For more information on the Akamai API, read the [property manager docs](https:/
...
+ routes[`/beta/apps/${APP_ID}`]       = { host: `http://localhost:${FRONTEND_PORT}` };
+ routes[`/apps/${APP_ID}`]            = { host: `http://localhost:${FRONTEND_PORT}` };
++routes[`/beta/config`]            = { host: `http://localhost:8889` };
+
+ module.exports = { routes };
+```
+
+Restart your insights-proxy to pick up the change.
+
+Create a `beta/config` directory inside of `cloud-services-config` and copy `main.yml` to it. Then, from the directory that the `beta` dir lives in, run `python -m SimpleHTTPServer 8889`. In your browser, go to `https://ci.foo.redhat.com:1337/beta/rhel/dashboard`. You should see something logged like this from SimpleHTTPServer:
```
kialam: > from the directory that the `beta` dir lives in

The wording here threw me for a bit as I was running the `SimpleHTTPServer ` from within `beta/config`. Suggestion: "from the `cloud-services-config` root directory run `python -m ...`"

kialam (approved):
I made one suggestion for clarification of where to serve the local `main.yaml` file, otherwise this worked great. Thanks!

README.md
```diff
@@ -132,3 +132,49 @@ For more information on the Akamai API, read the [property manager docs](https:/
 ## Build Process
 
 This repository has a webhook that automatically builds a [Jenkins job](https://jenkins-jenkins.5a9f.insights-dev.openshiftapps.com/job/akamai-config-deployer/) on every push. To configure this webhook, check the project's [webhook settings](https://github.com/RedHatInsights/cloud-services-config/settings/hooks)
+
+## Testing your changes locally
+
+Testing local changes is straightforward. First, add a line like this to your insights-proxy spandx config:
```
josejulio, line 138: Just a quick comment, following the [quick start guide](https://cloud.redhat.com/docs/storybook?path=/story/welcome--quick-start-guide) uses the file `config/spandx.config.js` instead of `profiles/local-frontend.js`.

I know that you are referring to: 
> your insights-proxy spandx config

but maybe it would be useful to use the defaults.

Note that I'm new here and maybe i'm wrong on assuming these are the defaults.

josejulio:
This information was really helpful, thanks!

README.md
ryelo: Good catch, I'm actually going to edit that to use the `local-frontend` -- which should be the config that is used.

README.md
```diff
@@ -132,3 +132,49 @@ For more information on the Akamai API, read the [property manager docs](https:/
...
+ routes[`/beta/apps/${APP_ID}`]       = { host: `http://localhost:${FRONTEND_PORT}` };
+ routes[`/apps/${APP_ID}`]            = { host: `http://localhost:${FRONTEND_PORT}` };
++routes[`/beta/config`]            = { host: `http://localhost:8889` };
+
+ module.exports = { routes };
+```
+
+Restart your insights-proxy to pick up the change.
+
+Create a `beta/config` directory inside of `cloud-services-config` and copy `main.yml` to it. Then, from the `cloud-services-config` dir, run `python -m SimpleHTTPServer 8889`. In your browser, go to `https://ci.foo.redhat.com:1337/beta/rhel/dashboard`. You should see something logged like this from SimpleHTTPServer:
```
karelhala: When I tried this command it threw `No module named SimpleHTTPServer` I would have to install this simple HTTP server, we simply can't asume that people understand python and have it properly setup.

Since all apps are using npm we should use node to serve this file. `npx http-server -p 8889` should do the same thing as `pythin -m SimpleHTTPServer 8889`

README.md
```diff
@@ -132,3 +132,49 @@ For more information on the Akamai API, read the [property manager docs](https:/
...
+
+ dashboard:
+-  title: Dashboard
++  title: XDashboardX
+   channel: '#flip-mode-squad'
+   deployment_repo: https://github.com/RedHatInsights/insights-dashboard-build
+   frontend:
+```
+
+Then, reload the site. You may not see your change at this point! Try clearing your local storage in your browser. To do this in Firefox, hit Shift-F9 and click "Local Storage", then right click on "https://ci.foo.redhat.com:1337" and delete all. Refresh the page and you should then see your changes. You'll notice too that SimpleHTTPServer logged another request. You will need to repeat this cache clearing step whenever you make changes to `main.yml` in your local enviornment.
```
karelhala: @panspagetka we might want to add `localStorage` value in our cache to make it store the value only for 5s (or less) for instance so devs doesn't have to clear their cache everytime they want to see their changes.

karelhala (changes requested):
Looking good! Thanks for adding this example, can you change python to `npx http-server` though?

README.md
PanSpagetka: Yeah, that sounds like a good idea! But it should be for dev setup only, right?

PanSpagetka: You have typo in environment (the last word of paragraph) ;)

README.md
beav: updated, thanks!

README.md
beav: typo fixed, thanks!

README.md
beav: marking as resolved, since change is happening in other doc

kruai (approved):
Really helpful addition, thanks!

README.md
karelhala: > Yeah, that sounds like a good idea! But it should be for dev setup only, right?

Well, users shouldn't play around with our debug functions. But if they's change it the only thing that would happen would be that they'd download `main.yml` multiple times in few minutes.

karelhala (approved):

PR #88, opened 2019-10-25
fix api docs path

ryelo:
adding a reload to the api-docs so that it directs to `docs/api`. If this isn't right, someone can feel free to commit to this branch as I'm going to be out


kruai:
@ryelo can you please remove the 2 extra `config.yml` files?

beta/config/main.yml
```diff
@@ -0,0 +1,450 @@
+ansible:
```
karelhala: I don't think you should push this file.

config/main.yml
```diff
@@ -0,0 +1,448 @@
+ansible:
```
karelhala: Or this one.

beta/config/main.yml
ryelo: whoops. I thought I had added these to my gitignore

ryelo:
@kruai @karelhala So now BU wants to remove product docs because some criteria changed. The path for api docs is still going to be docs/api, but I'm removing the product docs from the bundle for now.

kruai (approved):

karelhala (approved):

PR #89, opened 2019-10-30
add ingress so it shows in api docs

ryelo:


karelhala (approved):
Looking good! This will show only in docs api and not in any nav.

kruai (approved):

PR #93, opened 2019-11-05
Move Subscriptions to beta

ryelo:
~On hold until Nov 18~

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -427,7 +427,7 @@ staging:
   top_level: true
 
 subscriptions:
-  title: Subscription Reporting
+  title: Subscription Watch
   api:
     versions:
       - v1
@@ -438,7 +438,27 @@ subscriptions:
   frontend:
     paths:
       - /subscriptions
+    sub_apps:
+      - id: subscriptions-rhel
+        default: true
+        title: rhel
   git_repo: https://github.com/RedHatInsights/curiosity-frontend
+
+subscriptions-rhel:
+  title: Red Hat Enterprise Linux
+  frontend:
+    sub_apps:
+      - id: all
+        title: All
+        default: true
+      - id: arm
+        title: ARM
+      - id: ibmpower
+        title: IBM Power
+      - id: ibmz
+        title: IBM Z systems
+      - id: x86
+        title: x86
 
 vulnerability:
   title: Vulnerability Management
--- a/main.yml
+++ b/main.yml
@@ -439,11 +439,11 @@ subscriptions:
     paths:
       - /subscriptions
     sub_apps:
-      - id: subscriptions-rhel
+      - id: rhel-sw
         default: true
   git_repo: https://github.com/RedHatInsights/curiosity-frontend
 
-subscriptions-rhel:
+rhel-sw:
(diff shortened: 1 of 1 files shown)


kruai (approved):

ryelo:
@cdcabrera Please review. Going to close the other PR

cdcabrera:
Ok, ran through a local proxy run of some of the updates using both @karelhala and @kruai recommendations. 

Indeed, I've got a few tweaks within the application routing itself that need to happen, but was still able to confirm, "yes it did let me walk through some rough views".

Nov-18-2019 13-38-39

Talked with @ryelo I'm going to add a commit to this PR

cdcabrera:
@ryelo looks like this line causes some unexpected behavior on my local proxy run
Screen Shot 2019-11-18 at 1 54 44 PM

Screen Shot 2019-11-18 at 1 48 41 PM


Removing it seems to enable the menu as a dropdown again and it starts functioning as in the GIF above

ryelo:
@cdcabrera Removed!

cdcabrera (commented):
This should all work, based on local testing. However, this'll end up having the route/path `subscriptions/subscriptions-rhel/[sub app]. Confirming that this is ok on my side... (that title working would have been awesome)

Was looking at using `subscriptions/redhatenterpriselinux/[sub app]` but I'm guessing as far as naming schemes go for IDs... `subscriptions-rhel` may fit more inline @karelhala @kruai @ryelo ?

main.yml
```diff
@@ -424,25 +424,41 @@ staging:
...
     isBeta: true
   channel: '#subscriptions'
   deployment_repo: https://github.com/RedHatInsights/curiosity-frontend-build
+  top_level: true
   frontend:
     paths:
-      - /staging/subscriptions
+      - /subscriptions
+    sub_apps:
+      - id: subscriptions-rhel
```
cdcabrera: Confirmed, looks like we're going to flip the ID over to `rhel-sw`. We came up with a naming scheme since we're also going to have an `Openshift` left nav item, so if we can re-ID towards
```
- id: rhel-sw
```

That would help us out

main.yml
```diff
@@ -424,25 +424,41 @@ staging:
...
+        default: true
   git_repo: https://github.com/RedHatInsights/curiosity-frontend
 
+subscriptions-rhel:
```
cdcabrera: Continuation of the ID update...
`rhel-sw:`

cdcabrera (changes requested):
@ryelo apologies. after talking it over with the team it was decided that we needed a naming scheme for the paths. As @karelhala pointed out it is unfortunate we have `rhel`. 

To that point, and since we're also going to be needing an `openshift` at some point we opted to just use a suffix similar to

- `rhel-sw`
- and in the future, when we expand, `openshift-sw`

cdcabrera (approved):
Perfect, thank you sir!

Can confirm it's all appearing as intended
  
Screen Shot 2019-11-18 at 3 33 37 PM

kruai (approved):

PR #94, opened 2019-11-06
Add Settings app

PanSpagetka:
Add Settings app and move sources and RBAC into it.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -349,7 +349,6 @@ rhel:
 settings:
   title: Settings
   top_level: true
-  isBeta: true
   deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
   frontend:
     sub_apps:
@@ -365,7 +364,6 @@ settings:
 
 applications:
   title: Applications
-  isBeta: true
   deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
   frontend:
     sub_apps:
--- a/main.yml
+++ b/main.yml
@@ -319,9 +319,7 @@ rbac:
   frontend:
     title: User Access Management
     paths:
-      - /cost-management/settings/rbac
-      - /ansible/settings/rbac
-    reload: settings/rbac
+      - /settings/rbac
 
 remediations:
   title: Remediations
@@ -352,8 +350,17 @@ settings:
   title: Cloud Services Settings
   frontend:
     sub_apps:
+      - id: ''
+        default: true
+        title: 'User Preferences'
       - id: hooks
+      - id: sources
+      - id: rbac
   top_level: true
+  isBeta: true
+  deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
+  paths:
+  - /settings
 
 ruledev:
   title: Rule Development
@@ -407,9 +414,7 @@ sources:
   frontend:
     title: Sources
     paths:
-      - /insights/settings/sources
-      - /ansible/settings/sources
-    reload: settings/sources
+      - /settings/sources
 
 storybook:
   channel: '#flip-mode-squad'
--- a/main.yml
+++ b/main.yml
@@ -347,12 +347,12 @@ rhel:
   top_level: true
 
 settings:
-  title: Cloud Services Settings
+  title: Settings
   frontend:
     sub_apps:
       - id: ''
         default: true
-        title: 'User Preferences'
+        title: 'General'
       - id: hooks
       - id: sources
       - id: rbac
@@ -360,7 +360,7 @@ settings:
   isBeta: true
   deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
   paths:
-  - /settings
+    - /settings
 
 ruledev:
   title: Rule Development
--- a/main.yml
+++ b/main.yml
@@ -348,6 +348,9 @@ rhel:
 
 settings:
   title: Settings
+  top_level: true
+  isBeta: true
+  deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -352,8 +350,17 @@ settings:
   title: Cloud Services Settings
   frontend:
     sub_apps:
+      - id: ''
+        default: true
+        title: 'User Preferences'
```
ryelo: I think the mocks had this as "General"

https://marvelapp.com/5a4j976/screen/63049476

main.yml
```diff
@@ -352,8 +350,17 @@ settings:
   title: Cloud Services Settings
```
ryelo: This should just be "Settings"

https://marvelapp.com/5a4j976/screen/63049476

main.yml
```diff
@@ -352,8 +350,17 @@ settings:
...
+        default: true
+        title: 'User Preferences'
       - id: hooks
+      - id: sources
+      - id: rbac
   top_level: true
+  isBeta: true
+  deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
+  paths:
+  - /settings
```
ryelo: Let's indent this

ryelo (changes requested):
Just going off the mocks, a couple little naming things.

I'll sync with Kendra to double check

main.yml
```diff
@@ -352,8 +350,17 @@ settings:
...
 
 ruledev:
```
kruai, line 375: I'm not sure why the `settings` app is above `ruledev`, but I can fix that after merge.

main.yml
kruai: To clarify, just the `- /settings` line needs to be indented here.

kruai (changes requested):
I agree with Ryan's comments.

main.yml
karelhala: There's going to be another path in newar future for settings app `/settings/app` I don't really know how to structure the ID here @kruai any ideas. The url has to be `/settings/${smething}` and the `something` should indicate that it's a app settings. ID with `app` sounds strange and someone reading this file could understand it as placeholder.

karelhala: We'll need another `- settings/${config}` path, where I don't really know how to name the `config` app. We'll be configurating application in tat section, so `app` would be nice, but that's not really a good application ID. @kruai any ideas?

PanSpagetka: We can call it for example `- settings/sites` so it the whole url will look like `settings/sites/insights`. What do you think?

ryelo: Let's wait to see what Kendra's mocks say, but before it was going to be 
`settings/applications/insights` (but I'm sure apps would be fine)

main.yml
```diff
@@ -349,11 +347,32 @@ rhel:
...
+        default: true
+        title: 'General'
       - id: hooks
+      - id: sources
+      - id: rbac
+      - id: sites
   top_level: true
+  isBeta: true
+  deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
+  paths:
```
kruai: I also just noticed that `paths` isn't under `frontend` (it should be)

main.yml
```diff
@@ -349,11 +347,32 @@ rhel:
...
+
+sites:
+  title: Applications
+  frontend:
+    sub_apps:
+      - id: insights
+        title: Insights
+  isBeta: true
+  deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
+  paths:
```
kruai: I also just noticed that `paths` isn't under `frontend` (it should be)

PanSpagetka:
@kruai @ryelo @karelhala Please re-review, indentation should be right and paths should match https://github.com/RedHatInsights/settings-frontend/pull/10  if there are request for different paths/titles we can figure it out in next PR.

kruai (approved):
Having 2 applications define the same deployment repos will waste some space and CPU cycles, but we can figure out how to refactor a bit later. Otherwise, I think this looks good.

karelhala (approved):

main.yml
```diff
@@ -349,11 +347,32 @@ rhel:
...
+      - /settings/applications
+
+applications:
+  title: Applications
+  isBeta: true
+  deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
+  frontend:
+    sub_apps:
+      - id: insights
+          title: Insights
```
ryelo: This line is too far indented

ryelo (changes requested):

main.yml
PanSpagetka: Damn...fixed!

main.yml
```diff
@@ -349,11 +347,32 @@ rhel:
   top_level: true
 
 settings:
-  title: Cloud Services Settings
+  title: Settings
+  top_level: true
+  isBeta: true
```
kruai: Sorry I noticed this a little late, but `isBeta` is a field that only exists under `api`. Check out how it's used in other apps, like Approval, for instance.

main.yml
```diff
@@ -349,11 +347,32 @@ rhel:
...
-  top_level: true
+      - id: sources
+      - id: rbac
+      - id: applications
+    paths:
+      - /settings/applications
+
+applications:
+  title: Applications
+  isBeta: true
```
kruai: Same issue here.

kruai (changes requested):
Sorry, but I noticed another field that's out of place!

kruai (approved):
Looks good now.

ryelo (approved):

PR #100, opened 2019-11-15
Update Subscription Reporting paths and name

cdcabrera:
## Description
- Updates the app name towards `Subscription watch`, and
- expands the routing

## Relates to
Parallel task to #93 

## Open questions/guidance
Have additional questions regarding the implementation of ...

1. Was unsure of the naming scheme associated with the tiered menu structure, please advise if there is a specific ID for `subscriptions-rhel` that would be preferred?
2. It's been requested that there should be a `Getting Started` and `Documentation` set of links, similar to the below. I'm unfamiliar with that process, is that a setting controlled from `main.yml`?
   - https://github.com/RedHatInsights/curiosity-frontend/issues/129
   - https://github.com/RedHatInsights/curiosity-frontend/issues/130

Screen Shot 2019-11-15 at 3 36 12 PM

@kruai @ryelo @karelhala

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -422,7 +422,7 @@ staging:
   top_level: true
 
 subscriptions:
-  title: Subscription Reporting
+  title: Subscription Watch
   api:
     versions:
       - v1
@@ -435,7 +435,31 @@ subscriptions:
   frontend:
     paths:
       - /staging/subscriptions
+    sub_apps:
+      - id: subscriptions-rhel
   git_repo: https://github.com/RedHatInsights/curiosity-frontend
+
+subscriptions-rhel:
+  title: Red Hat Enterprise Linux
+  frontend:
+    sub_apps:
+      - id: all
+        title: All
+        reload: rhel/all
+        default: true
+      - id: arm
+        title: ARM
+        reload: rhel/arm
+      - id: ibmpower
+        title: IBM Power
+        reload: rhel/ibmpower
+      - id: ibmz
+        title: IBM Z systems
+        reload: rhel/ibmz
+      - id: x86
+        title: x86
+        reload: rhel/x86
+    suppress_id: true
 
 vulnerability:
   title: Vulnerability Management


main.yml
```diff
@@ -435,8 +435,32 @@ subscriptions:
...
+      - id: subscriptions-rhel
   git_repo: https://github.com/RedHatInsights/curiosity-frontend
 
+subscriptions-rhel:
+  title: Red Hat Enterprise Linux
+  frontend:
+    sub_apps:
+      - id: all
+        title: All
+        reload: rhel/all
```
karelhala, line 448: With reload you are loosing SPA experience and each nav click is refreshed, unless you want to change the URL I wouldn't add it here. Also `rhel` is quite unlucky because we already have rhel application. How exactly do you want your URL to look like? And how your nav is going to look like?

main.yml
```diff
@@ -422,7 +422,7 @@ staging:
   top_level: true
 
 subscriptions:
-  title: Subscription Reporting
+  title: Subscription watch
```
kruai: We've established title case as a standard for the sidenav in our applications; are you sure you want to change this to sentence case?

main.yml
```diff
@@ -435,8 +435,32 @@ subscriptions:
...
+      - id: ibmpower
+        title: IBM Power
+        reload: rhel/ibmpower
+      - id: ibmz
+        title: IBM Z systems
+        reload: rhel/ibmz
+      - id: x86
+        title: x86
+        reload: rhel/x86
+    suppress_id: true
```
kruai, line 462: I made this field to create a kind of exception for the "settings" sidenav item, since it wasn't an existing app that we still wanted to show. I'd like to avoid using this if possible, but I guess it depends on what you're trying to do with these sub-apps

kruai (commented):
There are a couple of things besides my comments that concern me in this PR:
- Since `subscriptions` is an app with a sub-app, and that sub-app has sub-apps, we're dealing with a new situation we haven't seen before (multiple levels of sub-apps in the sidenav). I'm pretty sure Chrome doesn't support that, but please correct me if I'm wrong @karelhala 
- As Karel said, those `reload` fields could lead to unexpected results since they start with `rhel`, and we already have the `rhel` bundle as a path. What URL do you want those sub-apps to have?

kruai:
Actually, disregard my first comment about the sub-app having sub-apps! I forgot that Ryan's PR is going to turn Subscriptions into its own bundle, so this shouldn't be an issue.

main.yml
cdcabrera: So this one has been confirmed, indeed, yes, the team wants to stick with `Subscription watch`, where the actual title would be all lower, but in the case of the left nav, they would like to continue with "sentence case"

@jeperry @bclarhk

ryelo: As per Allen Ng & Dan Oz this morning, we're all clear to use `Subscription Watch` on cloud.redhat.com, so we won't be using sentence case.

main.yml
cdcabrera: #### Regarding the path/route question:
Indeed, we noticed the heavy refresh on navigation. The original application path we were aiming for would based on our route param, so everything would follow similar to

-`[domain]/beta/subscriptions/rhel/?variant`, variant being the "sub-apps" so in the case of `all`
- `[domain]/beta/subscriptions/rhel/all` or optional, and then continuing in that style for the remaining `sub-apps`

- Specifically we were having trouble getting the routing to work when we removed the `reload`, so if there's an alternative approach we're up for it
- Plus, if we have a recommendation for something other than `rhel` I don't think the team would be opposed to altering that path/route

@kruai @karelhala

cdcabrera: Catching up with the team regarding `rhel` alternatives

cdcabrera: Ok @kruai @karelhala going to switch over to using `rhel` spelled out `redhatenterpriselinux`, seems to be the most direct alteration. Updating the PR

main.yml
cdcabrera: Can do, I'll go ahead and update it @ryelo

main.yml
cdcabrera: Going back through and looking at removing `supress_id` along with @karelhala recommendation of avoiding `reload`

main.yml
karelhala: Since you already know the "bundle" your app is going to live, it might be good to introduce it as well. You are currently using `staging` which is for temporary apps untill they know where they are going to be included and such.

So your navigation can look like:
```
[Home] | Subscription Watch
--------------------------------------
* Red Hat Enterprise Linux
  * All
  * ARM
  * IBM Power
  * IBM Z systems
  * x86
```

If that is the case we have to introduce new bundle and use this in it. Creating new bundle is fairly easy and can be done here, you can check how this was done for `Settings` https://github.com/RedHatInsights/cloud-services-config/pull/94/files

ryelo:
Closing in favor of #93. Please direct feedback to that PR.

PR #102, opened 2019-12-10
add limited settings

ryelo:
We only want to expose settings/applications/insights for now. Other apps will be following later.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -270,6 +270,28 @@ settings-ansible:
       - id: rbac
     suppress_id: true
 
+settings:
+  title: Settings
+  top_level: true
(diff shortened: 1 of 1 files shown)


kruai (approved):

ryelo:
Good to merge on 13 December

kruai:
@ryelo Can you update this PR with the newer structure of Settings that we have on prod-beta? That should allow it to be merged to Stable without issue

kruai (changes requested):
Explicitly marking as "Request changes" so this doesn't get merged before the config gets updated.

PR #106, opened 2020-01-10
Add sub-apps for compliance

dLobatog:
We were asked to split the app into Reports, Policies and Systems. However Policies will not be available in QA and prod for a while, how should we work around that?

This PR introduces "policies" - https://github.com/RedHatInsights/compliance-frontend/pull/197

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -104,6 +104,7 @@ compliance:
   api:
     versions:
       - v1
+    isBeta: true
   deployment_repo: https://github.com/RedHatInsights/compliance-frontend-build
   frontend:
     paths:
--- a/main.yml
+++ b/main.yml
@@ -110,16 +110,13 @@ compliance:
     paths:
       - /rhel/compliance
     sub_apps:
-      - id: ''
+      - id:
         title: Reports
         default: true
-        reload: compliance
       - id: policies
         title: Policies
-        reload: compliance/policies
       - id: systems
         title: Systems
-        reload: compliance/systems
 
 config:
   title: Cloud Services Config
--- a/main.yml
+++ b/main.yml
@@ -110,7 +110,7 @@ compliance:
     paths:
       - /rhel/compliance
     sub_apps:
-      - id:
+      - id: ''
         title: Reports
         default: true
       - id: policies
--- a/main.yml
+++ b/main.yml
@@ -108,6 +108,17 @@ compliance:
   frontend:
     paths:
       - /rhel/compliance
+    sub_apps:
+      - id: ''
+        title: Reports
+        default: true
+        reload: compliance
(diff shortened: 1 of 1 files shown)


ryelo:
@dLobatog when you say
> However Policies will not be available in QA and prod for a while, how should we work around that?

Do you mean policies won't be in qa-stable and prod-stable?

dLobatog (commented):
@ryelo That's right - it will sit on CI only for a while, as we want to add a few things to it before it can be deployed to other envs. It won't be useful at all in prod yet (changes to the client need to be made for that), so we don't want it there yet in order to avoid chances of breaking prod.

ryelo:
That's fine. When we push this out to the stable environments, we can leave the policies out.

main.yml
```diff
@@ -108,6 +108,17 @@ compliance:
   frontend:
     paths:
       - /rhel/compliance
+    sub_apps:
+      - id: ''
+        title: Reports
+        default: true
+        reload: compliance
```
kruai: Has this been tested locally? I haven't seen a `reload` defined for a default app before. If it was tested and working, I'll go ahead and approve this

dLobatog: Yeah, it works well - I removed it though and it works the same, to be fair I just took the `drift` app as an example and thought it was necessary.

dLobatog (commented):
It seems to work well as you can see in:

Screenshot from 2020-01-20 09-53-55

However if I click around the vertical nav in beta, for some reason it seems to point to the stable URL. Any idea what's going on there?

Peek 2020-01-20 09-52

Navigation in compliance-frontend is changed in beta through https://github.com/RedHatInsights/compliance-frontend/pull/287

kruai (approved):

ryelo:
@dLobatog It's probably your app loading the stable version of chrome versus the beta version of chrome

akofink (approved):
Looks good. :+1:

kruai:
@ryelo If you think this looks good as well, I'll go ahead and merge!

main.yml
```diff
@@ -104,10 +104,19 @@ compliance:
   api:
     versions:
       - v1
+    isBeta: true
   deployment_repo: https://github.com/RedHatInsights/compliance-frontend-build
   frontend:
     paths:
       - /rhel/compliance
+    sub_apps:
+      - id:
```
kruai: This should be `- id: ''` if you want the ID to be blank.

kruai (changes requested):
Actually, there's one small change that needs to be made before merge.

ryelo (approved):

kruai (approved):

PR #112, opened 2020-01-22
Cost management is now using platform sources

chambridge:
* No need for a cost management specific sources left nav
* Platform sources will be accessed via the settings "gear" in the top nav


karelhala (approved):
Nice! This is realy good news.

PR #116, opened 2020-02-21
If user is not entitled to insights hide it in settings

karelhala:


kruai (approved):
Ohhhh interesting! I would assume that these new vars are used in Chrome. Very neat :)

karelhala:
@kruai yes, exactly. This is basically to hide nav items based on some information chrome holds.

PR #117, opened 2020-02-26
Permissions args is an array, so use them as array

karelhala:


README.md
```diff
@@ -142,7 +142,7 @@ If you want to hide any navigational element based on some chrome's logic, this
 
 #### permissions.args
 
-If method requires some arguments in order to properly work this is the way how to pass them to it.
+If method requires some arguments in order to properly work this is the way how to pass them to it - and array of items.
```
kruai: Can we change it to the following for readability?
`If the permissions method requires some arguments in order to properly work, this is how to pass them to it: an array of items.`
If "the permissions method" is wrong, feel free to change it; I'm not sure what "If method" is referring to on that line.

kruai (changes requested):

README.md
karelhala: Yeah, this is correct, it refers to `permissions.method`, I'll change it to 
```
If the `permissions.method` requires some arguments in order to properly work, this is how to pass them to it: an array of items.
```

kruai (approved):

PR #125, opened 2020-03-10
Add apiName to advisor API so we generate the doc URL correctly

karelhala:
The API frontend work is done here https://github.com/RedHatInsights/api-frontend/pull/93 both PRs can be merged separately

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -3,6 +3,7 @@ advisor:
   api:
     versions:
       - v1
+    apiName: insights
   channel: '#advisor'
   deployment_repo: https://github.com/RedHatInsights/insights-advisor-frontend-build
   frontend:


kruai (commented):
This is fine with me, though... is the "versions" field still used? Since I assume that field is used to build the API's URL as well, do you think we could just combine them into something like this?
```
api:
  paths:
    - /insights/v1
```
Or, since there only seems to be one version per API, maybe just:
```
api:
  path: /insights/v1
```

karelhala:
Yup, the api version is used to build out the full URL. We could (and probably shloud) revisit the API properties. But that would require rewriting a bunch of code in CSC, we just have to coordinate and probably ask folks around what would be the most logical and best approach to this.

kruai (approved):

PR #127, opened 2020-03-10
Update Settings and Sources configs in Stable

kruai:
Takes what we have for Sources/Settings in our Beta envs and puts it in ci-stable.


ryelo (approved):

ryelo:
This is good to merge anytime today. Just let me know so I can push other things out too.

PR #129, opened 2020-03-11
update ci-stable based on feedback

ryelo:


kruai (changes requested):
- If you remove `general`, you should also get rid of the `settings-general` app.
- If `rbac` is now the default, you'll want to add `/settings` as a route for that app.

kruai (approved):

PR #131, opened 2020-03-11
update csc for prod-stable push (add settings)

ryelo:
On hold until March 12


kruai (approved):
Looks like what's in ci-stable, so I approve!

ryelo:
@kruai looks like I removed a nav item from the ansible bundle by accident. I added it back and the diff is now correct!

PR #137, opened 2020-03-19
Removing custom-policies from settings/application

josejulio:
We moved to user-preferences


josejulio:
@karelhala please review :)

karelhala (approved):

PR #144, opened 2020-03-26
Only Org Admins can change Cost Management Application Settings

chambridge:
We want only entitled org_admin users to be able to alter cost management settings.

@karelhala Can you take a look and verify that my change works as described?

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -50,6 +50,7 @@ applications:
           method: isEntitled
           args:
             - cost_management
+          method: isOrgAdmin
     paths:
       - /settings/applications
 


kruai (approved):
I'm approving bc this won't affect any other apps, but I'll wait for @karelhala to give his approval before we merge.

karelhala:
This is not correct way, as you have duplicate keys in permissions object. We can't have 2 methods per one permission (enhancement is being considered), however you can add `permissions` to `cost-management` app itself
```yml
cost-management:
  permissions:
    method: isOrgAdmin
  title: Cost Management
  top_level: true
  api:
    versions:
      - v1
    isBeta: true
  channel: '#koku'
  deployment_repo: https://github.com/RedHatInsights/cost-management-build
  description: >
    Cost Management is an application that provides users a view of their
    enterprise cost. Cost Management allows users to evaluate the cost on
    infrastructures like Amazon Web Services (AWS) across multiple *master*
    accounts, breaking down to service usage, compute usage, and storage
    allocation. Details can also be broken down utilizing tagging to view
    data attributed to project or cost center assignments. Cost Management
    also allows users to breakdown cost at the platform level by associating
    cost with OpenShift clusters, nodes, and the projects running their. Cost
    can be determined by deriving it from the underlying infrastructure (like
    AWS) or deriving it based on a user set of rates.
  docs: https://koku.readthedocs.io/en/latest/
  frontend:
    paths:
      - /cost-management
    sub_apps:
      - id: ''
        title: Overview
        default: true
        group: cost-management
      - id: details
        group: cost-management
      - id: cost-models
        title: Cost models
        group: cost-management
  git_repo: https://github.com/project-koku/
  mailing_list: cost-mgmt@redhat.com
```

karelhala (changes requested):
yml doesn't allow same keys in one object

chambridge:
@karelhala We don't want OrgAdmin on the entire Cost Mgmt application, just on who can change the settings.

So it sounds to me like we need to change the permission check to only be isOrgAdmin, but this setting will now show up for customers that don't have access to Cost Management.

karelhala:
@chambridge oh I see! Then you can add `permissions` checker for settings app as well
```yml
applications:
  title: Applications
  deployment_repo: https://github.com/RedHatInsights/settings-frontend-build
  permissions:
    method: isOrgAdmin
    apps:
      - cost-management
  frontend:
    sub_apps:
      - id: insights
        title: Insights
        permissions:
          method: isEntitled
          args:
            - insights
      - id: cost-management
        title: Cost Management
        permissions:
          method: isEntitled
          args:
            - cost_management
    paths:
      - /settings/applications
```

Looks like I forgot to add this to config documentation, I'll add it there so you and others can properly use this!

PR #150, opened 2020-03-30
Show Cost Mgmt Settings for entitled org admin

chambridge:
Make @karelhala update to support app settings for cost management only visible for entitled org admin users:
https://github.com/RedHatInsights/cloud-services-config/pull/144#issuecomment-605990375


karelhala (approved):
That's it! Looking good

kruai (approved):
I think I'll start adding @karelhala as a reviewer when the `permissions` field is included :)

PR #166, opened 2020-04-10
Cost Management details links not highlighted

dlabrecq:
The Cost Management details links are not highlighted properly.

To see the issue, navigate to Cost Management > Details > Infrastructure

1.  Select Azure from the tertiary nav. The Details > Infrastructure link is no longer highlighted.
2. While still on Azure, refresh the page. Now the Details > OpenShift link is highlighted.

This small change fixes both issues.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -204,10 +204,9 @@ details:
     sub_apps:
       - id: ocp
         title: OpenShift
-        group: cost-management
       - id: aws
+        default: true
         title: Infrastructure
-        group: cost-management
 
 policies:
   title: Policies


karelhala (approved):

karelhala:
Isn't this going to cause refreshes when navigating from `/cost-management/` to `/cost-management/deetails/ocp` if you see nav updates when navigating in your app then you might be calling `insights.chrome.appNavClick`

dlabrecq:
I'm not observing any refresh issues when `group` is omitted. The page is landing properly with your koku-ui fix.

Should I leave `group` in place and file a separate issue to address the highlight?

I would like to keep the `default` prop, which fixes issue number 2.

karelhala:
Oh, nice! We can go with this approach if it's not causing any issues.

PR #188, opened 2020-04-21
Migration Analytics in Production

carlosthe19916:
Migration Analytics will go to Production and I'd like to merge this PR for having our URL into Production.

- Just one question: is it OK to keep `title: Experimental API` [1]? or what should be the correct value?

[1] https://github.com/carlosthe19916/cloud-services-config/blob/patch-5/main.yml#L310

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -290,6 +290,31 @@ landing:
   channel: '#flip-mode-squad'
   deployment_repo: https://github.com/RedHatInsights/landing-page-frontend-build
 
+migrations:
+  title: Migrations
+  frontend:
+    sub_apps:
+      - id: migration-analytics
+        default: true
+  top_level: true
+
+migration-analytics:
+  title: Migration Analytics
+  api:
+    versions:
+      - v1
+    alias:
+      - xavier
+    tags:
+      - value: "experimental"
+        title: "Experimental API"
+  deployment_repo: https://github.com/RedHatInsights/xavier-ui-deploy
+  frontend:
+    paths:
+      - /migrations
+      - /migrations/migration-analytics
+  git_repo: https://github.com/project-xavier/xavier-ui
+
 openshift:
   title: OpenShift (OCM)
   api:


kruai:
@carlosthe19916 my understanding is that this should become active on 4/23; do you know what time we should aim for?
I'm also not sure what the tags fields are for. Do you know more @karelhala @ryelo ?

kruai (approved):

ryelo:
Not sure about the experimental API, but 4/23 is the date we have so hold this until then.

main.yml
```diff
@@ -290,6 +290,31 @@ landing:
...
+      - id: migration-analytics
+        default: true
+  top_level: true
+
+migration-analytics:
+  title: Migration Analytics
+  api:
+    versions:
+      - v1
+    alias:
```
karelhala, line 306: This means that in the api-docs page we'll show the app as xavier instead of `Migration Analytics` is that correct?

main.yml
```diff
@@ -290,6 +290,31 @@ landing:
...
+  title: Migrations
+  frontend:
+    sub_apps:
+      - id: migration-analytics
+        default: true
+  top_level: true
+
+migration-analytics:
+  title: Migration Analytics
+  api:
```
karelhala, line 303: I tried looking for `migration-analytics` on the `/api` andpoint and it looks like there is no such api, but I could find `/api/xavier` do you mind adding
```yaml
  api:
    ...
    apiName: xavier
    ...
```

main.yml
```diff
@@ -290,6 +290,31 @@ landing:
...
+    versions:
+      - v1
+    alias:
+      - xavier
+    tags:
+      - value: "experimental"
```
karelhala, line 309: Tags currently does not have any effect, but good point in adding these! You can keep these in here, it will just have no effect, do you want to somehow show in the UI that users might experience some problems when accessing this API?

karelhala (changes requested):
Other than `apiName` I am fine with this, few questions but overall looks good!

PhilipCattanach:
@ryelo @kruai 
Hi Asa/Ryan,

This is the first time my Engineering team has deployed a new application to Insights Production so we may need some help/direction from yourselves.

We had hoped to deploy the prod stable branch today. With the intention of allowing the Engineering team and QA team to be able to do some testing to ensure all is well using the URL https://cloud.redhat.com/migrations/migration-analytics.

Then on Thursday (23rd April) the Insights homepage will be updated with the new Migration Services panel with the link to the Migration Analytics URL.

Is that something that can be done? Or do both steps need to be done on the 23rd? 

Thanks,
Phil

karelhala:
@PhilipCattanach it can be merged prior 23rd if you want to test things out. It will be accessible on the URL you posted, but since there is no direct link in he UI to this application I am fine with merging it users won't be able to access it unless they know the URL.

kruai:
I mean, we do have pre-production environments as well... it would be great if we could do QA testing in the QA environment. Why do we need to do QA testing on Production for this app?

PhilipCattanach:
@kruai 
Asa, we have tested the application in QA already. It is that we are being very 'risk adverse' for our first deployment to prod-stable.

main.yml
carlosthe19916: We'd like to show `Migration Analytics` instead of `xavier`. I'll add a new commit to change it. Thanks

PhilipCattanach:
@karelhala 
Karel - exactly, you get the point - it will only be the Engineering team and QE who will have access to the application prior 23rd.

main.yml
carlosthe19916: @karelhala I'll add it no problem. I just wonder why the other branches worked fine without `apiName: xavier`? I mean, our config for `ci-stable, qa-stable, etc` did not include `apiName: xavier`

main.yml
PhilipCattanach: Thanks for the feedback. This PR has the same changes we merged into the qa-stable branch. We will apply the changes that you have suggested starting from ci-stable to validate and promote them.

main.yml
PhilipCattanach: As above

main.yml
carlosthe19916: let's keep it since it will have no effect... showing messages like `users might experience some problems when accessing this API` were not in our plans so I think it is OK for now not to show such message

PhilipCattanach: This value is fine for this PR. 
Looking ahead - can you provide a list of the other permissible values please so we can decide which tag is most appropriate for Migration Analytics?

kruai:
OK, if nobody broadcasts the URL and it isn't added to the landing page, I'm fine with it! Once Karel's comments have been addressed, I think this will be fine to merge today.

main.yml
karelhala: It totaly got out of my radar, currently the xavier API is not listed on API docs because there is missing openAPI spec but with this applied and if you'd were to expose openapi.json on `/api/xavier/v1/openapi.json` customers could check your API and use it directly without using app UI.

main.yml
karelhala: Yeah, sorry that I didn't noticed this before, but it totally got out of my radar.

main.yml
karelhala: Currently we are not showing any tags in api docs, but there were plans to show them. We were planning to show just the value and title so users can filter the table in docs api.

main.yml
PhilipCattanach: @kruai 
Asa - Sorry for the confusion. Myself and Carlos were updating in parallel. Just to reiterate we won't be making a change to this PR. We must release to prod-stable the same configuration that we have tested in ci-stable and qa-stable.

main.yml
PhilipCattanach: @kruai 
Asa - Sorry for the confusion. Myself and Carlos were updating in parallel. Just to reiterate we won't be making a change to this PR. We must release to prod-stable the same configuration that we have tested in ci-stable and qa-stable.

PhilipCattanach:
@kruai 
Asa - Are you satisfied with the replies to all of Karel's comments? Can you proceed and apply the PR?

main.yml
karelhala: Ah, okay. Sure thing! If you want to discuss the docs api feel free to ping me and we can discuss it more deeply.

karelhala (approved):

kruai:
Yep, that's fine with me. Merging shortly.

PR #212, opened 2020-05-19
Adds notifications and automation calculator to analytics

benthomasson:


kruai:
This is fine, but I just want to note that I don't see `automation-calculator` in some of the other environemnts. You may want to open additional PRs to make your environments match

PR #218, opened 2020-05-26
Adds Systems sub-app to advisor

AllenBW:
in support of https://projects.engineering.redhat.com/browse/RHCLOUD-6431

not yet ready to be merged, need the advisor pieces to be in place as well, n then we'll have to coordinate

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -12,6 +12,8 @@ advisor:
     sub_apps:
       - id: recommendations
         title: Recommendations
+      - id: systems
+        title: Systems
       - id: topics
         title: Topics
   source_repo: https://github.com/RedHatInsights/insights-advisor-frontend


main.yml
```diff
@@ -12,6 +12,8 @@ advisor:
     sub_apps:
       - id: recommendations
         title: Recommendations
+      - id: systems
```
AllenBW, line 15: @kruai gotta question about https://github.com/RedHatInsights/insights-advisor-frontend/pull/653/files#diff-14b1e33d5bf5649597cdc0e4f684daddR13

we're doing `systems(redirect) { insights.chrome.appNavClick({ id: 'systems', redirect }); }` as is, without this pr merged, when navigating to /advisor/systems the nav item that's selected is Compliance -> Systems, wondering if advisor needs to add any additional info to the nav click to ensure the correct selection is made

kruai: I wouldn't think so, but you're free to [test this change locally](https://github.com/RedHatInsights/cloud-services-config#testing-your-changes-locally) to make sure it works before I approve it!

AllenBW: @kruai so this is what happens when clicking on the nav item.... they both expand 😱 

Screen Shot 2020-05-26 at 1 02 55 PM

shdunning: @AllenBW what if you give this a unique id, like `advisor_systems`?

AllenBW: @shdunning issue with that is nav url redirect is built off the id, which means we'd have to use the path `/advisor_systems` in advisor (rather than `/systems` ) 👎 
Screen Shot 2020-05-27 at 12 35 54 PM

shdunning: oh rite - bummer :-(

kruai: Yeah, that's really unfortunate... @karelhala is there any special config I could have CSC provide to Chrome to mitigate this issue? Perhaps something like the `group` tag?

AllenBW: https://projects.engineering.redhat.com/browse/RHCLOUD-6708 is the jira for work fixin this...

PR #233, opened 2020-07-13
adds "my user access" page and gates RBAC on orgAdmin

ryelo:
Just want an ack from @karelhala to double check everything here before merge

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -344,6 +344,16 @@ migration-analytics:
       - /migrations/migration-analytics
   source_repo: https://github.com/project-xavier/xavier-ui
 
+my-user-access:
+  title: My user access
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -344,6 +344,16 @@ migration-analytics:
       - /migrations/migration-analytics
   source_repo: https://github.com/project-xavier/xavier-ui
 
+my-user-access:
+  title: My user access
+  channel: '#platform-experience'
+  deployment_repo: https://github.com/RedHatInsights/insights-rbac-ui
```
kruai: The deployment repo should probably be different than the source repo, unless this is a weird one

kruai (approved):

PR #241, opened 2020-07-29
Added CCX in ci-beta

quarckster:


kruai (changes requested):
I have a few questions:
- Can you please move this config so that it's between Catalog and Chrome? Keeping this list alphabetized will help keep it human-readable
- What's the purpose of this config? Entries generally have repos and paths defined. Is this getting consumed by another service?

quarckster:
@kruai Sure, I'll move it. Wrt purpose. I spoke about it with @karelhala. Our UI lives under OCM but we have our own API. It would be logical to create a sub_app under `openshift` but I was told that OCM team doesn't publish their spec, therefore, docs for our app won't be shown. That's why I put the config for app into a separate stanza.

karelhala:
@kruai as @quarckster said, this config file is used in api documentation and since CCX does not have proper code repository this is the minimal config to expose it in the table.

Anyway we had a discussion with @ryelo the other day on cleaning CSC main.yml to be just navigation and some other files to include other global configurations (API docs, user configuration, etc.). But there are no final words or anything on it.

kruai:
OK, just wanted to make sure this config was going to do what you wanted before we merge it. Once you move it to its alphabetically-correct position, I'll merge it :)

karelhala (approved):

kruai (approved):

PR #250, opened 2020-08-03
[qa-beta] Introduce api subitems and use it in drift application (#238)

karelhala:


kruai (approved):
Thanks for adding the info to the README as well! Doesn't look like it'll break current integrations, so I think it's good to go.

PR #261, opened 2020-08-28
Adds trust frontend

AllenBW:
on /security/insights


karelhala (approved):
Are the `tilte`s correct? It will have bundle name `Insights Security` and in navigation there will be `Insights Security`, also shouldn't these be sentence case?

But overall nothing seems to be off.

AllenBW:
@karelhala good call, looks like sentence case is correct, and we don't so much need a title override, so getting rid of that second one 👍

karelhala:
Just to be sure, this is not used in any bundle as of right now, right? So inaccessible by users.

kruai (approved):

kruai:
Everything looks good to me, too. Merging!

PR #269, opened 2020-09-29
Introduce api subitems and use it in drift application (#238)

karelhala:


README.md
```diff
@@ -15,6 +15,11 @@ Here is some example configuration that demonstrates the structure, using all re
         versions:
```
kruai, line 15: This `versions` no longer exists in your new structure, right?

kruai: Oh, but I guess it makes sense to keep this until all apps move to this new format

README.md
```diff
@@ -62,6 +67,14 @@ The following properties are used if your app has an API:
 
 This is the list of API versions your app can use. Since `v1` is the default, you'll usually want at least that one defined.
 
+#### app_id.api.subItems
+
+If your API consists of multiple APIs you can list them in here it has the same signature as `{app_id.api}`.
```
kruai: Grammatical change:
"If your API consists of multiple APIs, you can list them here. Each has the same signature as `{app_id.api}`."

README.md
```diff
@@ -62,6 +67,14 @@ The following properties are used if your app has an API:
...
+
+#### app_id.api.apiName
+
+If your API is accessible on oher URL than `/api/{app_id}/{versions[0]}/openapi.json` you can change this by passing correct name, the URL will look like `/api/{apiName}/{versions[0]}/openapi.json`
```
kruai: Grammatical change:
"If your API is accessible on a URL other than `/api/{app_id}/{versions[0]}/openapi.json`, you can change it by passing the correct name. The URL will look like `/api/{apiName}/{versions[0]}/openapi.json`."

kruai (changes requested):

README.md
karelhala: Thank you! Should be fixed.

README.md
karelhala: Thank you! Should be fixed.

README.md
karelhala: Well apps can choose from two ways of defining their API docs:
* Flat - most apps will use this since they are using just one API docs
* Nested - if app uses multiple API docs they can define them in nested structure

kruai: Ah okay, that makes sense!

kruai (approved):

PR #295, opened 2020-10-22
fix casing for sap

ryelo:


kruai:
I'm cherrypicking this to qa-stable as well, since you made one for both ci-stable and prod-stable

PR #318, opened 2020-11-27
Add Strimzi UI

b1zzu:


b1zzu:
cc @pmuir

pmuir:
I just want to check that this won't cause the app to appear in the menus...

b1zzu:
@pmuir any update? can we add it in this way without it appearing in the menus?

pmuir:
Sorry, bad english - I mean - "before we merge this, are we sure that it won't cause the app to appear in the menus... I don't think it will but a maintainer should confirm.

kruai:
@pmuir @b1zzu Adding this won't flat-out add it to any menus. You'll need to add its ID to another app's `sub_apps` for that

PR #322, opened 2020-12-01
chore: add rhosak control plane UI

pmuir:


main.yml
```diff
@@ -477,6 +477,13 @@ rhosak-data-plane-ui-build:
     paths:
       - /application-services/rhosak-data-plane-ui-build
 
+rhosak-control-plane-ui-build:
+  title: Control Plane UI for RHOSAK
+  deployment_repo: https://github.com/RedHatInsights/rhosak-control-plane-ui-build
+  frontend:
+    paths:
+      - /application-services/rhosak-control-plane-ui-build
```
b1zzu: @pmuir I think we can remove the `frontend` object also for the control-plane-ui because it will be used as a module for the dashboard right :thinking:  
```suggestion
```

pmuir: Done.

kruai: If he removes `frontend`, the code in `deployment_repo` won't be deployed. What do you mean by it being used as a module for the dashboard?

kruai (approved):

PR #346, opened 2021-01-19
add ROS to main.yml

kgaikwad:


main.yml
```diff
@@ -741,3 +742,19 @@ vulnerability:
       - id: systems
         title: Systems
   source_repo: https://github.com/RedHatInsights/vulnerability-ui
+
+ros:
```
kruai: I want to keep this config as alphabetized as possible. Can you please move this new config to be between `rhel-sw` and `ruledev`?

kruai (changes requested):

main.yml
kgaikwad: surely, fixed the order and indentation as well.

main.yml
```diff
@@ -604,6 +605,22 @@ rhel-sw:
       - id: x86
         title: x86
 
+ros:
+  title: Resource Optimization
+  api:
+    versions:
```
kgaikwad: @kruai, as of now there is no API version released so will it be possible to add changes without this `versions` value? or is it mandatory to have?

main.yml
```diff
@@ -604,6 +605,21 @@ rhel-sw:
       - id: x86
         title: x86
 
+ros:
+  title: Resource Optimization
+  api:
```
karelhala: You don't have to include the api part at all if it's not ready. The API team should share api spec on `/${YOUR_APP_NAME}/${API_VERSION}/openapi.json` to allow customers to work with it. If it's not ready yet, you don't have to add it here and wait for it to be ready, once that you can open new PR to include it.

kgaikwad: Great! As per a suggestion, removed api section. Thanks @karelhala.

main.yml
kruai: Looks like Karel answered this for you below, so I'm resolving :)

kruai (approved):

PR #367, opened 2021-02-03
chore(ci-beta): add cloud mosaic odh tutorials

redallen:
cc @dgutride


kruai:
@redallen What are your expectations for `mosaic`? You'll need to set `top_level` if you want it to be its own app group (like `insights`, `cost-management`, `ansible`, etc.

redallen:
@kruai I just expect to be able to host some static files for POC federated module work, but `top_level` would be nice so I added it!

kruai (approved):

PR #368, opened 2021-02-03
Attempt to make the entire bundle serve from a single

pmuir:


kruai (changes requested):
@pmuir The "openshift-streams" and "quickstarts" apps currently use different deployment repos, but in your PR, they're both using https://github.com/RedHatInsights/rhosak-dashboard-build. Is this intentional, and you're no longer using https://github.com/RedHatInsights/rhoas-guides-build?

pmuir:
Yes - we are using the repos like rhoas-guides-build solely to present the JS content on the /apps/<app-name> path, and not to provide the index.html. The content is then assembled using webpack module federation inside the host app.

IOW what I want is all the menu items pointing to the dashboard app, then the react router in that takes care of pointing to the right federated module.

So it sounds like those sub app paths will all end up pointing to the dashboard app...

kruai (approved):
Oh, okay - If you now have your project set up so that the react router handles that path, and you don't need the other repo anymore, then I think this change should be fine. Just give me the heads up when you've tested it on your end and are ready for me to merge this

PR #392, opened 2021-02-11
Use complex module definition for edge and api-docs

karelhala:
#### No app loading for edge and api-docs

Currently no app is being loaded in edge or api-docs because of first level navigation. This PR changes the module definition for these 2 apps so it uses complex module definition so we can use all partials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ To indicate which module should be loaded when rendering your app (you can have 
 
 ##### app_id.frontend.module.group
 
-If you have first level application to indicate which group should be managed by this module.
+If you have a first-level application, this field indicates which group should be managed by this module.
 
 #### app_id.frontend.paths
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -111,6 +111,10 @@ To indicate federated modules scope of your application (you can have multiple s
 
 To indicate which module should be loaded when rendering your app (you can have multiple modules per one scope). This is usually `./RootApp`
 
+##### app_id.frontend.module.group
+
+If you have first level application to indicate which group should be managed by this module.
```
kruai:
```suggestion
If you have a first-level application, this field indicates which group should be managed by this module.
```

main.yml
```diff
@@ -707,9 +714,6 @@ rhel-sw:
 
 ros:
   title: Resource Optimization
-  api:
-    versions:
-      - v0
```
kruai: Was this change intentional?

kruai (changes requested):

main.yml
karelhala: Good catch!

kruai (approved):

PR #395, opened 2021-02-12
Add modules to main.yml and add validation

karelhala:
### Add modules

If someone is running chrome and migrated app locally they could experience problems with missing module config. This PR adds the module definitions to each of migrated app, so once the chrome lands in the ci-stable all migrated apps will start to use it.

### Add validation

Since we want to have validation on ci-stable same as is on ci-beta this PR adds package.json and the required configs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ Here is some example configuration that demonstrates the structure, using all re
     deployment_repo: https://github.com/app-deployment-repo-url
     disabled_on_prod: true
     docs: https://link.to.docs.com/docs
+    permissions:
+        method: isOrgAdmin
+        apps:
+            - app_id_1
     frontend:
         title: App Title Override
         paths:
@@ -35,6 +39,10 @@ Here is some example configuration that demonstrates the structure, using all re
             -   id: app_id_1
                 title: Some Sub App
                 default: true
+                permissions:
+                    method: isEntitled
+                    args:
+                        - insights
             -   id: app_id_2
                 title: Another Sub App
         suppress_id: true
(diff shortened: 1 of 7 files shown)


Hyperkid123 (approved):

README.md
```diff
@@ -69,11 +77,11 @@ This is the list of API versions your app can use. Since `v1` is the default, yo
 
 #### app_id.api.subItems
 
-If your API consists of multiple APIs, you can list them here. Each has the same signature as `{app_id.api}`.
+If your API consists of multiple APIs you can list them in here it has the same signature as `{app_id.api}`.
 
 #### app_id.api.apiName
 
-If your API is accessible on a URL other than `/api/{app_id}/{versions[0]}/openapi.json` you can change it by passing the correct name. The URL will look like `/api/{apiName}/{versions[0]}/openapi.json`
+If your API is accessible on oher URL than `/api/{app_id}/{versions[0]}/openapi.json` you can change this by passing correct name, the URL will look like `/api/{apiName}/{versions[0]}/openapi.json`
```
kruai: I think the "before" version is better than the "after" version here. Strangely, this change would make it match with ci-beta. So was it OK in ci-beta first, and then changed to a less correct version? :thinking:

main.yml
```diff
@@ -19,7 +20,7 @@ advisor:
   source_repo: https://github.com/RedHatInsights/insights-advisor-frontend
 
 ansible:
-  title: Red Hat Ansible Automation Platform
+  title: Ansible Automation Platform
```
kruai, line 23: There are lots of additional changes in this PR. For instance, multiple changes to the titles of apps. Were these intentional?

main.yml
```diff
@@ -348,7 +355,7 @@ ingress:
 insights:
   title: Insights
   frontend:
-    title: Red Hat Insights
+    title: Insights
```
kruai, line 363: Title changed here as well

main.yml
```diff
@@ -423,7 +431,7 @@ openshift:
       - v1
   deployment_repo: https://github.com/RedHatInsights/uhc-portal-frontend-deploy.git
   frontend:
-    title: Red Hat OpenShift Cluster Manager
+    title: OpenShift Cluster Manager
```
kruai, line 439: One more

kruai (changes requested):
Had a few notes on this one. Not sure what to do about the README changes... it looks like ci-beta's README needs to be updated to match some of them for this one, tbh.

README.md
kruai: OK, I've updated this part of the README on all envs, so you can discard these 2 changes

kruai: OK, I've updated this part of the README on all envs, so you can discard these 2 changes

main.yml
karelhala: Yup, they are intentional we don't want to show Red Hat anymore, if you want I can open new PR just for these changes.

README.md
karelhala: Sure thing!

main.yml
kruai: No it's OK! Just wanted to make sure it was intentional.

kruai (approved):

PR #402, opened 2021-02-16
Add reviewers to CODEOWNERS

kruai:
I've been trying to transition me off of Fakamai and Akamai work ever since I joined HBI back in September. You two touch the front-end way more than I do, so I think it makes sense for you to be aware of new PRs to the repo as well. Is that all right with you, and is there anybody else I should add to this?


kruai:
FYI @gmcculloug ^

ryelo (approved):

PR #406, opened 2021-02-16
add module for cost management

redallen:
Related: https://github.com/project-koku/koku-ui/pull/1857

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -268,6 +268,10 @@ cost-management:
   frontend:
     paths:
       - /cost-management
+    module:
+      appName: cost-management
+      scope: costManagement
+      module: ./RootApp
     sub_apps:
       - id: ''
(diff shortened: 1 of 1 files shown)


dlabrecq (approved):

main.yml
```diff
@@ -268,6 +268,10 @@ cost-management:
   frontend:
     paths:
       - /cost-management
+    module:
+      appName: cost-management
+      scope: costManagement
+      module: ./RootApp
```
karelhala, line 275: Cost management is a mix of first level and second level navigation. You should add `group: cost-management` in here in order to use correct module if user refreshes page with for instance `/beta/cost-management/ocp` without it the UI would break with error from component.

karelhala (approved):

PR #413, opened 2021-02-18
fix(starter): add starter app to ci-beta

redallen:
Add the starter app so that forks of https://github.com/RedHatInsights/frontend-starter-app don't have to change `main.yml` before testing their apps.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -457,6 +457,7 @@ insights:
       - id: remediations
       - id: registration
       - id: ros
+      - id: starter
   top_level: true
 
 integrations:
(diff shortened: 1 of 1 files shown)


kruai:
Is your intention to have it show up on the sidenav for Insights applications? If not, I'd recommend putting it in the `sub_apps` section of the [Staging bundle](https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/main.yml#L803) instead

redallen:
Yes that's my intention @kruai . I think it's nice for devs who fork the repo but don't edit `main.yml` to see what the chrome's side nav might look like.

karelhala:
@redallen yeah I think it would be reasonable to add the starter app to the staging section. It would require some more changes to starter app, but we don't want to show starter app in the navigation for insights bundle, at least not before we shoot the recordings for summit.

redallen:
What environment are you going to shoot recording against @karelhala ? I'm happy to target any environment.

kruai:
@redallen If you take `- id: starter` out of Insights's sub-apps list and add it to [Staging's](https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/main.yml#L803) sub-apps instead, I think this will be fine in ci-beta, qa-beta, etc.

kruai (approved):
Looks good now!

PR #422, opened 2021-02-23
Openshift nav changes for Summit 2021

PanSpagetka:
https://issues.redhat.com/browse/RHCLOUD-12315

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -590,14 +590,54 @@ openshift:
         default: true
       - id: subscriptions
         title: Subscriptions
+        sub_apps:
+          - id: openshift-sw
+            title: OpenShift
+            group: subscriptions
+            reload: ../subscriptions/openshift-sw
         group: openshift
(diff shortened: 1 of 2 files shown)


main.yml
```diff
@@ -581,16 +581,22 @@ openshift:
     paths:
       - /openshift
     sub_apps:
+      - id: overview
```
karelhala, line 585: The id should be `''` otherwise it would break the openshift flow.

karelhala (approved):

karelhala (changes requested):
The id in overview of OCM should be `''`

main.yml
karelhala: Oh, the overview just moved up! Let's see what will do this reoder in the URL and navigation

karelhala (approved):

karelhala (approved):

PR #437, opened 2021-03-08
Subscription watch openshift update

cdcabrera:
Add in for Subs Watch views OCP and Dedicated

@rblackbu @bclarhk

PR #438, opened 2021-03-08
Subscription watch openshift update

cdcabrera:
Add in for Subs Watch views OCP and Dedicated

@rblackbu @bclarhk

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -620,7 +620,12 @@ subscriptions:
         default: true
         group: subscriptions
       - id: openshift-sw
-        title: Red Hat OpenShift
+        title: OpenShift
+        sub_apps:
+          - id: ocp
+            title: Container Platform
+          - id: dedicated
+            title: Dedicated
         group: subscriptions
   source_repo: https://github.com/RedHatInsights/curiosity-frontend
 


kruai (changes requested):
You can't nest `sub-apps` on an item in `sub-apps`; you'll need to separate that out into its own application config (unless that functionality was added without my knowledge @karelhala @ryelo ?)
Also changes should go into ci-stable first, before they're considered for qa-stable.

cdcabrera:
@kruai it appears nesting is allowed in CI-beta, this works... nesting does not work in CI-stable the test caught it. We had assumed the test was in place on QA-stable. 

- This one confirms nesting is available in CI-beta, https://github.com/RedHatInsights/cloud-services-config/commit/6bb9940b2c1ecac0652ba67a58ccf7c9737fad42
- This one fails testing for nesting, https://github.com/RedHatInsights/cloud-services-config/pull/437

Just need to confirm 
- what environment the ability to nest stops at? 
- and/or is there a planned date to support nesting in the higher environments?

Off to the side we can go ahead and plan to break the item out, we'll wait for response to your question @kruai

kruai:
@cdcabrera Ah, you're right. I forgot that this was a top-level app and that we had built in the ability to nest sub-apps for those. Sorry for the confusion! It will still need to go into ci-stable before qa-stable, but I no longer have an issue with the structure

cdcabrera:
Closing this down per discussion with @karelhala see RHCLOUD-12690

PR #439, opened 2021-03-09
Add new application-services nav

pmuir:
New nav for app services

Note that there is a bug at the moment which I reported to @ryelo and @Hyperkid123 earlier today. We should get that sorted before merging this one.


pmuir:
I also have a patch to our bundle code to route all these paths, so coordinating deployment would be good.

pmuir:
See #444

PR #446, opened 2021-03-12
Update subscription navigation to match new mocks

karelhala:
This PR removes the `rhel-sw` from navigation and replaces it with `subscriptions` which is the wanted scenario. It also introduces `section: insights` to OCM navgiation and adds cost-management and subscription to it.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -617,7 +617,7 @@ openshift:
               title: Container Platform
             - id: openshift-dedicated
               title: Dedicated
-            - id: opneshift-sw-documentation
+            - id: openshift-sw-documentation
               title: Documentation
               navigate: https://access.redhat.com/documentation/en-us/cost_management_service/
       - id: cost-management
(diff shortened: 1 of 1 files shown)


PanSpagetka (approved):

main.yml
```diff
@@ -611,23 +611,18 @@ openshift:
...
-          - id: opneshift-sw-documentation
-            title: Documentation
-            group: subscriptions
-            navigate: https://access.redhat.com/documentation/en-us/cost_management_service/
-        group: openshift
+            - id: openshift-container
+              title: Container Platform
+            - id: openshift-dedicated
+              title: Dedicated
+            - id: opneshift-sw-documentation
```
kruai: I know you didn't change this line, but this would be a great time to fix the `opneshift` typo :)
```suggestion
            - id: openshift-sw-documentation
```

PR #448, opened 2021-03-13
Flatten Cost Management nav (ci-beta)

dlabrecq:
As pre-work to moving Cost Management under OpenShift, the links shown in the Insights navigation pane must be flattened, first.

This ensures the Insights nav matches the UI's updated routing paths.

https://issues.redhat.com/browse/COST-1145

Screen Shot 2021-03-12 at 10 47 16 PM

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -294,8 +294,38 @@ cost-management:
             args:
               - url: '/api/cost-management/v1/user-access/?type=OCP'
                 accessor: 'data'
-      - id: infrastructure
+      - id: aws
+        title: Amazon Web Services
         group: cost-management
+        permissions:
+          - method: apiRequest
+            args:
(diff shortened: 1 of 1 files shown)


karelhala:
@dlabrecq was this discussed with PMs? Because we have a bit different navigation direction - https://issues.redhat.com/browse/RHCLOUD-12315

dlabrecq:
Yes and we also spoke with Ryan. The team thought Cost Management's routing paths should match the flattened nav first. 

I know we will be under OpenShift, eventually -- this will help ensure the UI and nav will work together when we're finally moved over.

ryelo (approved):
Yeah this is just an in-between step. Just want to make sure everything is working when flattened and then when moved into their correct location it should act the same.

kruai (approved):
LGTM

PR #462, opened 2021-03-23
Add Subscription Watch

joeldmyers:
Adds Subscription Watch to main.yml.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -982,7 +982,7 @@ subscriptions:
       - id: subscriptions-documentation
         title: Subscriptions Documentation
         navigate: https://access.redhat.com/products/subscription-central?extIdCarryOver=true&sc_cid=701f2000001Css5AAC
-      - id: manifests
+      - id: subscription-central
         title: Manifests
 
       
--- a/main.yml
+++ b/main.yml
@@ -935,6 +935,20 @@ storybook:
       - /docs/storybook
   source_repo: https://github.com/RedHatInsights/insights-frontend-storybook
 
+subscription-central:
+  title: Subscription Central
+  channel: '#subscription-central'
+  deployment_repo: https://github.com/RedHatInsights/subscription-central-ui-build
+  frontend:
+    module:
+      appName: subscription-central
+      scope: subscriptionCentral
+      module: ./RootApp
+      group: insights
+    paths:
+      - /insights/subscriptions/manifests
+  source_repo: https://github.com/RedHatInsights/subscription-central-ui
+
 subscriptions:
   title: Subscription Watch
   api:
--- a/main.yml
+++ b/main.yml
@@ -936,7 +936,7 @@ storybook:
   source_repo: https://github.com/RedHatInsights/insights-frontend-storybook
 
 subscription-central:
-  title: Subscription Central
+  title: Manifests
   channel: '#subscription-central'
   deployment_repo: https://github.com/RedHatInsights/subscription-central-ui-build
   frontend:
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -935,6 +935,20 @@ storybook:
       - /docs/storybook
   source_repo: https://github.com/RedHatInsights/insights-frontend-storybook
 
+subscription-central:
```
karelhala, line 938: This should be `manifests` so the subs navigates user to this application.

main.yml
```diff
@@ -935,6 +935,20 @@ storybook:
...
+  title: Subscription Central
+  channel: '#subscription-central'
+  deployment_repo: https://github.com/RedHatInsights/subscription-central-ui-build
+  frontend:
+    module:
+      appName: subscription-central
+      scope: subscriptionCentral
+      module: ./RootApp
+      group: insights
```
karelhala: No need for this group as it's a helper to idenitfy the app within flat first leven nav.

main.yml
```diff
@@ -935,6 +935,20 @@ storybook:
...
+    paths:
+      - /insights/subscriptions/manifests
```
karelhala, line 948: It's missing a `reload` function in here so the app is correctly loaded
```suggestion
    paths:
      - /insights/subscriptions/manifests
    reload: subscriptions/manifests
```

main.yml
karelhala: Does the app have API? If so it's good to include the api part as well so the docs are autogenerated from the openapi.json spec.

karelhala (commented):
Just a few adjustments and we are golden!

main.yml
ryelo: @karelhala would you rather use manifests here? I added their identifier in their UI as `subscription-central`

joeldmyers: Subscription-central may be better, since long-term (from my understanding) this intends to have more than just manifests.  But let me know if that creates an issue with routing.

The app itself is just a front-end; it gets data from RHSM-API - https://api.access.redhat.com/management/v1/allocations; does that need to be registered in this?

main.yml
joeldmyers: ok I added that in.

main.yml
joeldmyers: ok removed.

main.yml
karelhala: You can keep the `subscription-central` as app name, however you'll have to change in the subs section https://github.com/RedHatInsights/cloud-services-config/blob/c4661b14d4c9ed88d3b44ad937b130bd8092ad82/main.yml#L985-L986

To be
```YAML
      - id: subscription-central
```

joeldmyers: I think the line that you highlighted is actually part of the Subscription Watch app's config, not this new Subscription Central app.  So I don't think we should change that.  Let me know if we should change anything within our config.

karelhala: Yup, that's subs watch app config, however that'll create nav url `insights/subscriptions/manifests` (in future `/rhel/subscriptions/manifests`) which should lead to this app, right?

joeldmyers: OK, yes that's correct; I had to go through docs to confirm this separate update was for us.  I just pushed an update for that.

karelhala: Nice! What do you want to have in the left navigation actually? `Manifests` or `Subscription Central`

joeldmyers: It should be Manifests.

karelhala: In that case you should remove the `title: Manifests` from Subs watch config and here change the `title` to `Manifests`.

joeldmyers: OK just updated that; thanks!

karelhala (approved):

PR #469, opened 2021-03-25
Update sed to connector

ryelo:


karelhala (approved):
Shouldn't the id change as well? If not, add `reload: settings/connector` so the correct URL is loaded.

ryelo:
@karelhala I'll update the id here and in the dev repo to connector

karelhala (approved):

PR #476, opened 2021-03-26
NOTIF-140 Attempt to fix config

josejulio:
- There is still an issue with the nav menu not selecting the right
   element

It was loading forever in `beta`, adding some `module` properties made it load correctly, but the selected menu is wrong.

Screenshot_2021-03-26_12-42-48

It should be `Notifications -> Insights`.

@karelhala @Hyperkid123 Do you have any idea what I'm missing?

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -570,6 +570,15 @@ my-user-access:
 notification:
   title: Notifications
   frontend:
+    app_base: user-preferences
+    module:
+      appName: notification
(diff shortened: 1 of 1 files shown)


karelhala (approved):
Yeah, the problem is that the nav treats this as seperate app, we should investigate mixing first level nav and second mixed together. Adding this task to post-summit features. However good fix @josejulio!

PR #477, opened 2021-03-29
Policies to chrome 2.0

rvsia:
cc @Hyperkid123 @karelhala

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -299,6 +299,7 @@ policies:
   channel: '#cloudservices-policies'
   deployment_repo: https://github.com/RedHatInsights/policies-ui-frontend-build
   frontend:
+    module: policies#./RootApp
     paths:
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -299,6 +299,8 @@ policies:
   channel: '#cloudservices-policies'
   deployment_repo: https://github.com/RedHatInsights/policies-ui-frontend-build
   frontend:
+    module: policies#./RootApp
+    section: security
```
karelhala: Sections are not supported on prod-beta. Should be just the module definition.

karelhala (changes requested):

main.yml
rvsia: upps forgot to add the commit

karelhala (approved):

PR #482, opened 2021-03-30
NOTIF-163 & NOTIF-167 Rename insights bundle and add openshift

theute:
Requires @josejulio validation.

https://github.com/RedHatInsights/notifications-backend/pull/153 would have to be merged soon after or things won't work


josejulio:
This also requires a change in user-preferences app. WIll do and link it back

josejulio:
https://github.com/RedHatInsights/user-preferences-frontend/pull/63 would have to be merged soon after as well.

PR #484, opened 2021-03-31
Simplify user-preferences navigation

karelhala:


main.yml
```diff
@@ -991,7 +975,17 @@ user-preferences:
       - id: email
         title: Email Preferences
         default: true
-      - id: notification
+        group: email
+      - id: notifications
+        title: Notifications
+        group: email
```
josejulio, line 981: Out of curiosity, why `email` ? Is it because it's the same group used in `user-preferences.frontend.module.group` ?

karelhala: Yup, the groupping of nav items is a bit off. We have a plan to refactor the config and make it more understandable.

PR #504, opened 2021-04-09
Fix some redirects based on feedback

astrozzc:
1. The roi-calculator should be automation-calculator based on feedback of automation-analytic team
they are not changing the name until after summit;
2. The new urls for cost-management cloud resources are without "infrastructure"
3. Add /ansible/automation-analytics to /ansible/insights
4. Nicole says they reverted the changes of /openshift/xxx yesterday and not plan for the summit


kruai (approved):
LGTM

PR #520, opened 2021-04-19
Add Releases page to OpenShift Cluster Manager

rhamilto:


karelhala (approved):
Looking good! @rhamilto the feature is not yet deployed to qaproaduth-beta environment, right? Can you ping us once it's there?

PR #529, opened 2021-04-21
update qa-stable

ryelo:


karelhala (approved):

main.yml
```diff
@@ -354,20 +392,21 @@ ingress:
...
-      - id: patch
-      - id: drift
       - id: policies
-      - id: inventory
-      - id: remediations
+      - id: patch
+      - id: subscriptions
+        section: business
       - id: registration
-      - id: ros
```
kgaikwad, line 370: I would like to know - is `ros` sub-app unintentionally removed from qa-stable or any reason *behind it?

karelhala: @ryelo probably took what is on `prod-beta` and mapped it to `qa-stable`. Which envs should have the `ros` application?

kgaikwad: It is not yet in prod-beta but `ros` application was present in CI/QA/Stage.

PR #537, opened 2021-04-22
Add back user-preferences/notifications

josejulio:


josejulio:
@ryelo this adds back some things we need for notifications in stage, thanks!

PR #541, opened 2021-04-23
Enable notifications/integrations on stable

josejulio:


ryelo:
@josejulio can we close this in favor of the batch update in #549

josejulio:
Changes already in stable

PR #586, opened 2021-05-03
ON HOLD: Adds malware app support

AllenBW:


karelhala (approved):

karelhala:
I can't seem to find the app defined anywhere in the file. Where is this app deployed and how do you access it? @AllenBW

AllenBW:
OOO does it have to be defined in the file? @karelhala  right now i can't locally dev on this cuz it gets replaced with the dashboard we also have it hosted somewhere internally... but not on crc at the moment

app does the chrome identification here... https://github.com/RedHatInsights/malware-detection-frontend/commit/22676de4f807cea55753224fe80cd3215a63e480

karelhala:
@AllenBW yup, it has to be hosted on CRC as well. Otherwise it won't show up. I am kinda suprised it woked before. Did it have navigation and such? For local dev this approach might work, but for accessing it on ci or on prod this won't work at all.

PR #611, opened 2021-05-10
image-builder: Migrate to chrome 2.0

croissanne:


main.yml
```diff
@@ -454,6 +458,8 @@ image-builder:
         matcher: isNotEmpty
   frontend:
     section: operations
+    module: image_builder#./RootApp
```
karelhala: The `image-builder` app has dash in the name. So we have to slightly tweak the `module` defiitin
```suggestion
    module:
      appName: image-builder
      scope: imageBuilder
      module: ./RootApp
```

karelhala:
@Gundersanne is the chrome 2 PR pushed to prod-beta? If so, you might want to open another PR for prod-beta branch of this config.

croissanne:
@karelhala currently prod-beta does not include the updated chrome 2 webpack configs. Trying to see if it works in staging first, then I'll make a release and open up a pr for prod-beta.

main.yml
croissanne: Right but I defined [the module](https://github.com/osbuild/image-builder-frontend/blob/main/config/prod.webpack.config.js) as `image_builder`, as a module name with a `-` wasn't supported. So it should be `image_builder` no?

karelhala: Oh, so the `appName` should be `image-builder` and `scope` should be `image_builder`

karelhala (approved):

karelhala:
@Gundersanne you'll have to remove the `skipChrome2` flag from the config and all should click together.

PR #620, opened 2021-05-14
feat(openshift): add federated module

redallen:
Now that uhc-portal is chrome 2 enabled the `main.yml` can be updated.

PR #633, opened 2021-05-19
feat(openshift): add federated module

redallen:
Syncing #620 to other envs.


redallen:
@karelhala I've confirmed with @elad661 that `uhc-portal` is using Chrome 2 in prod, so this is safe to merge.

karelhala (approved):

PR #650, opened 2021-05-29
[ci-beta] Set the Ansible Dashboard as the default path for the ansible path

lgalis:
Set the Ansible Dashboard as the default path for the ansible path


lgalis:
@karelhala, @ryelo - please review

karelhala (approved):

PR #653, opened 2021-06-01
[ci-beta] Remove the isBeta flag for catalog and approval apis

lgalis:
Remove the isBeta flag for catalog and approval apis so they will be displayed in the doc api list


lgalis:
@karelhala, @ryelo  - please review

karelhala (approved):

PR #659, opened 2021-06-02
Add ansible-dashboard group to ansible-dashboard

karelhala:
There is an error on ci/qa/stage-beta where if user navigates to non-chrome2 app in ansible bundle dashboard is always visible. This PR fixes such issue, by using different group for the ansible-dashboard (other than the section partial) and using this group definition in app definition as well.


Hyperkid123:
cc @lgalis

Hyperkid123 (approved):

PR #662, opened 2021-06-03
NOTIF-217 Adding application services

theute:
@josejulio could you have a look ?
I didn't create the bundle/app/event-type yet but will do.
Is there anything else to setup for this new bundle + app + event-type ?

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -588,6 +588,9 @@ notifications:
       - id: openshift
         title: OpenShift
         group: notifications
+      - id: application-services
+        title: Application Services
+        group: notifications
   source_repo: https://github.com/RedHatInsights/notifications-frontend
 
 openshift:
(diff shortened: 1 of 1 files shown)


karelhala (approved):

karelhala:
@theute the settings part looks good! You might want to add `application-services` config to [RedHatInsights/user-preferences-frontend~src/config/config.json#L31-L35](https://github.com/RedHatInsights/user-preferences-frontend/blob/master/src/config/config.json#L31-L35) if `notifications` application is going to serve the DDF schema and handle the requests.
```JSON
"application-services": {
  "title": "Application services",
  "application": "notifications",
  "resourceType": "notification-preference?bundleName=application-services"
}
```

We have a plan to refactor the navigation, in this effort we'll try to merge these 2 configs together so we won't have to add it on two places.

EDIT: I'll wait for the follow up PR on user-preferences so we won't end up with broken links.

josejulio:
> @josejulio could you have a look ?
> I didn't create the bundle/app/event-type yet but will do.
> Is there anything else to setup for this new bundle + app + event-type ?

As @karelhala mentioned, the user-preferences PR - I'll create it later today.

josejulio:
The PR is here: https://github.com/RedHatInsights/user-preferences-frontend/pull/73

PR #666, opened 2021-06-07
Chrome nav rework

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-14525

### TO DO
- [x] port main left nav to new definition
- [x] add landing page schema
- [x] add settings schema
- [x] add user preferences schema
- [x] add module definition file
- [x] document new format
- [x] add schema validator
- [x] split `navigation.json` into multiple files
- [x] add routes specification to the `fed-modules.json` (possibly rename it)
- [x] cleanup?
- [x] make sure the new schemas are hosted at `/config`
- [x] integrate https://github.com/RedHatInsights/cloud-services-config/pull/668


chrome/navigation.json
```diff
@@ -0,0 +1,736 @@
...
+            "title": "Red Hat Enterprise Linux",
+            "pathname": "/insights"
+        },
+        {
+            "id": "ansbile",
+            "title": "Ansible Automation Platform",
+            "pathname": "/ansible"
+        }
+    ],
+    "userPrefferences": [
```
ryelo:
```suggestion
    "userPreferences": [
```

chrome/fed-modules.json
```diff
@@ -0,0 +1,90 @@
+[
+    {
+        "moduleScope": "userPreferences",
+        "manifestLocation": "/apps/user-preferences/fed-mods.json"
```
karelhala: I know we want it to be as permissive as possible, but a lot of these have default values. Might be good to have this only if it differs from `/apps/${moduleScope}/fed-mods.json`

chrome/navigation.json
```diff
@@ -0,0 +1,736 @@
+{
+    "landingPage": [
+        {
+            "id": "application-services",
+            "title": "Application services",
+            "pathname": "/application-services"
```
karelhala: I'm kinda leaning towards same approach as with module defintion. Unless the `pathname` differes from `id` use the `id` field. At least for the landing page nav.

chrome/navigation.json
```diff
@@ -0,0 +1,736 @@
...
+            "title": "Red Hat Enterprise Linux",
+            "pathname": "/insights"
+        },
+        {
+            "id": "ansbile",
+            "title": "Ansible Automation Platform",
+            "pathname": "/ansible"
+        }
+    ],
+    "userPreferences": [
```
karelhala: Is there any reason why `userPreferences` and `settings` is lifted from `leftNav`? I understand the `landingPage` navigation. But if we are speaking just about navigation we shouldn't treat these any differently than other parts.

Speaking about the content, there should be a special file just to populate `userPreferences` content (mostly copy of what is already in the [config.json](https://github.com/RedHatInsights/user-preferences-frontend/blob/master/src/config/config.json))

chrome/navigation.json
```diff
@@ -0,0 +1,736 @@
...
+                        "method": "isEntitled",
+                        "args": [
+                            "cost_management"
+                        ]
+                    }
+                }
+            ]
+        }
+    ],
+    "leftNav": [
```
karelhala: I think that it would be much better to split these defintions to each file
* `settings.json`
* `rhel.json`
* `ansible.json`
* `openshift.json`
* `edge.json`
* `user-preferences.json`
* `application-services.json`

And just either keep mapping here or in chrome. Both ways have something good in them
* in chrome
  * `+` just one call
  * `-` breaking the simplicity of defining nav just in here
* in here
 * `+` nav config in here
 * `-` multiple calls to fetch nav

chrome/navigation.json
```diff
@@ -0,0 +1,736 @@
...
+            "id": "application-services",
+            "title": "Application services",
+            "pathname": "/application-services",
+            "navItems": []
+        },
+        {
+            "id": "openshift",
+            "title": "Openshift",
+            "pathname": "/openshift",
+            "navItems": [
```
karelhala: I would change this to object instead of array
```JSON
"navItems": {
  "/openshift": {},
  "/openshift/overview": {},
  "/openshift/releases": {},
  "/openshift/subscriptions/openshift-container": {
    "subPathOf": {
        "groupId": "insights",
        "title": "Insights",
    },
    "id": "openshift-subscribtions-containers",
    "title": "Container Platform",
  },
  "/openshift/quota": {
    "subPathOf": {
        "groupId": "insights",
        "title": "Insights",
    }
  },
  "/openshift/subscriptions/openshift-dedicated": {
    "subPathOf": {
        "groupId": "insights",
        "title": "Insights",
    }
  }
}
```

We'll also have to figure out how to pull `openshift` code when user navigates to `["/openshift", "/openshift/overview", "/openshift/releases", "/openshift/quota"]`

chrome/navigation.json
Hyperkid123: Yes we want do this.

chrome/navigation.json
Hyperkid123: Stuff will be plit into different files.

chrome/fed-modules.json
Hyperkid123: We talked about the fed modules files and we will change it so it will be a Routes definition instead. Looking somehing like this:
```json
{
    "dashboard": {
        "manifestLocation": "/apps/dashboard/fed-mods.json",
        "modules": [
            {
                "id": "rhel-dashboard",
                "module": "./RootApp",
                "pathname": [
                    "/insights", 
                    "/insights/dashboard"
                ]
            },
            {
                "id": "ansible-dashboard",
                "module": "./RootApp",
                "pathname": [
                    "/ansible",
                    "/insights/dashboard"
                ]
            }
        ]
    },
}
```

This will allow us to easily identify which module to use and each module will have its own `Route` so there won't be any conflicts between different modules.

chrome/navigation.json
Hyperkid123: We will use `appId` to identify to which app the link leads. `appId` will be based on the fed module ID (the insights.appname from package.json)

chrome/navigation.json
Hyperkid123: After some discussions, we won't be doing this. It will be again hard to construct the actual tree of the nav. We will lift the module definitions and routes definition from the navigation schema.

The navigation schema will be purely a mirror of the DOM and after splitting the schema into multiple files (based on namespace) there is no benefit for doing this.

Hyperkid123:
@coderbydesign I've modified the Jenkins file to copy any JSON files from `/chrome` dir to `/chrome` dir in akamai. This is where we would like to store the new schemas.

Jenkinsfile
```diff
@@ -106,6 +106,7 @@ node {
           cp \"$AKAMAI_HOST_KEY\" ~/.ssh/known_hosts
           chmod 600 ~/.ssh/known_hosts
           rsync -arv -e \"ssh -2\" *.yml sshacs@cloud-unprotected.upload.akamai.com:${AKAMAI_APP_PATH}
+          rsync -arv -e \"ssh -2\" ./chrome/*.json sshacs@cloud-unprotected.upload.akamai.com:${AKAMAI_APP_PATH}/chrome
```
john-dupuy, line 110: @Hyperkid123 I see that we will be still be uploading the main yml file. Is the plan to support that for some deprecation period and then eventually get rid of it? 

Having a deprecation period would give me time to update the smoke tests for the new split config.

john-dupuy: Furthermore, I think we need to backup these JSON files in an analogous way to https://github.com/RedHatInsights/cloud-services-config/pull/666/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R46
So that they can be re-uploaded as the main.yml/release.yml are in https://github.com/RedHatInsights/cloud-services-config/pull/666/files#diff-e6ffa5dc854b843b3ee3c3c28f8eae2f436c2df2b1ca299cca1fa5982e390cf8R125-R128

Hyperkid123: Currently, the `main.yml` is still required for the akamai/fakamai to host some static asses. Those will be required until every app migrated over chrome 2.

We don't know when it's gonna happen but we will start pushing the migration a bit more now. Hopefully, I will be able to give you a precise date soon.

Hyperkid123:
@ryelo @karelhala @john-dupuy OK I think it's ready.

john-dupuy (approved):

Hyperkid123:
Merging this, there are no breaking changes to current stuff. I assume we will have to do a couple of updates in the next few days/week.

PR #668, opened 2021-06-10
Adds ocm-advisor

AllenBW:
fixes https://issues.redhat.com/browse/CCXDEV-5083

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -620,6 +620,8 @@ openshift:
       - id: releases
         title: Releases
         group: openshift
+      - id: ocp-advisor
+        group: insights
       - id: subscriptions
         title: Subscriptions
         section: insights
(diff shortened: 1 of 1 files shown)


AllenBW:
cc @gkarat

main.yml
```diff
@@ -644,6 +646,20 @@ openshift:
...
 
+ocp-advisor: 
+  title: Advisor
+  source_repo: https://github.com/RedHatInsights/ocp-advisor-frontend
+  deployment_repo: https://github.com/RedHatInsights/ocp-advisor-frontend-build
+  frontend:
+    module: ocp-advisor#./RootApp
+    isHidden: true
+    paths:
+      - /openshift/advisor
```
gkarat: There was a requirement for the new Advisor to run under the `/openshift/insights/advisor` URL, so it seems we need to rewrite it to the suggested format :)

main.yml
```diff
@@ -644,6 +646,20 @@ openshift:
...
+    sub_apps:
```
gkarat, line 658: Could you also add the `clusters`, please?

main.yml
```diff
@@ -644,6 +646,23 @@ openshift:
 
   top_level: true
 
+ocp-advisor: 
+  title: Advisor
+  source_repo: https://github.com/RedHatInsights/ocp-advisor-frontend
+  deployment_repo: https://github.com/RedHatInsights/ocp-advisor-frontend-build
+  frontend:
+    module: ocp-advisor#./RootApp
+    isHidden: true
```
gkarat, line 655: Just a question to make sure: when `isHidden` is set to true, then the app is not available in all envs within its path, right?

AllenBW: @gkarat ishidden ensures the entry does not show up in the lnav this pr only adds a csc record for ci-beta

gkarat:
LGTM! 🎉  As far as I can see

Hyperkid123 (approved):

PR #680, opened 2021-06-16
Update new chrome schemas

Hyperkid123:
### Changes
- Add missing link in settings sections
- FIx automation hub navigation links
- Fix federated module naming (do not use `-` in the name)

### DIFF
--- a/chrome/ansible-navigation.json
+++ b/chrome/ansible-navigation.json
@@ -3,7 +3,7 @@
     "title": "Ansible Automation Platform",
     "navItems": [
         {
-            "appId": "ansible-dashboard",
+            "appId": "ansibleDashboard",
             "title": "Overview",
             "href": "/ansible/ansible-dashboard"
         },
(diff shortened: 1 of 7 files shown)


chrome/ansible-navigation.json
```diff
@@ -3,7 +3,7 @@
     "title": "Ansible Automation Platform",
     "navItems": [
         {
-            "appId": "ansible-dashboard",
```
rvsia, line 6: Aren't `appId`s used somewhere inside applications? For example, to register the app on mount.

Hyperkid123: App ID is used to determine what kind of link we then use. We have to pick between react-router link and the normal `a` tag. If we navigate to or from an application that does not use federated modules we need a full page refresh. So if a current appId does not match the link appId and the module is not dynamic we use the normal `a` tag.

These schemas have absolutely nothing to do with rendering or initializing. It's just for navigation.

rvsia (approved):

PR #705, opened 2021-06-24
Fix config

astrozzc:
1. Criteria in prod should use prod host name;
2. In prod, use stage gateway secret for FedRAMP origin because we still use stage gov cluster
3. Remove top level redirect for gov.xxx


coderbydesign (approved):
This looks to cover the issues we tested manually earlier against Akamai staging for the cloud.redhat.com property. Thanks, Jay!

ryelo (approved):

PR #696, opened 2021-06-23
Add new Downloads page to OCM left navigation

cben:
- [x] ~~Don't merge yet, the version currently in production https://cloud.redhat.com/openshift/downloads is older but we should get a good-to-launch version after today's OCM UI deploy~~...

https://issues.redhat.com/browse/SDA-4211

PR #718, opened 2021-06-28
[qa-beta] Add new Downloads page to OCM left navigation

cben:
https://qaprodauth.cloud.redhat.com/beta/openshift/downloads is a new page, adding here similar to #696.
(https://issues.redhat.com/browse/SDA-4211)

@karelhala @Hyperkid123 I'm not sure whether this PR is right/needed, as README claims there is automatic ci-beta -> nightly-beta -> qa-beta -> stage-beta sync, but I see nightly-beta and qa-beta haven't updated recently?
[This is not urgent at all, we almost never use qaprodauth/beta :see_no_evil:]

Also, I only now noticed that ci-beta had both main.yml and openshift-navigation.json.  In #696 I only updated the latter, here updating both.  Should I go back and update main.yml on ci-beta first?

### DIFF
--- a/chrome/openshift-navigation.json
+++ b/chrome/openshift-navigation.json
@@ -18,6 +18,11 @@
             "title": "Releases"
         },
         {
+            "appId": "openshift",
+            "href": "/openshift/downloads",
+            "title": "Downloads"
+        },
+        {
             "groupId": "insights",
             "title": "Insights",
             "navItems": [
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@cben the branches should be synced but we have some issue in jenkins so we may have to do it manually.

And yes we will be switching to a new format that is only on CI beta right now (We will be sending information about the changes soon). If you can update the JSON the do it. Otherwise, I will take care of it.

Hyperkid123 (approved):

PR #719, opened 2021-06-28
add product ids

ryelo:
Ideally, we should use these for case creation on each page. In chrome, we'll either use the value provided here, or empty if there is no product.

### DIFF
--- a/chrome/ansible-navigation.json
+++ b/chrome/ansible-navigation.json
@@ -14,27 +14,32 @@
                 {
                     "appId": "ansibleAutomationHub",
                     "title": "Collections",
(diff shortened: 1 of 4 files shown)


Hyperkid123 (approved):

himdel:
~~This PR broke the menus for Ansible Hub on qa and stage~~, cc @ZitaNemeckova @newswangerd 

~~(Our routes under `/ansible-hub/*` do NOT start with `ansible-`.)~~

Sorry, wrong PR, this doesn't add the prefixes. It's https://github.com/RedHatInsights/cloud-services-config/commit/f54b0159a9d212ba97ba46d4fd5e9f0dda2739af

PR #735, opened 2021-07-06
add cloud tutorials to nav, rename rhel-navigation to insights-navigation

redallen:
Also update docs per https://github.com/RedHatInsights/insights-chrome/pull/1505


chrome/validate-nav.js
```diff
@@ -6,16 +6,16 @@ const navigationSchema = require('./validationSchemas/navigation');
 
 
 
-const navigationFiles = ['ansible', 'application-services', 'rhel', 'settings', 'user-preferences', 'openshift'].map((file) => path.resolve(__dirname, `${file}-navigation.json`))
+const navigationFiles = fs.readdirSync(__dirname)
+  .filter(f => f.endsWith('-navigation.json'))
+  .filter(f => f !== 'landing-navigation.json');
```
karelhala: Two extra iterations over the same array. I'd merge it into one `.filter(f => f !== 'landing-navigation.json' && f.endsWith('-navigation.json'))`

chrome/validate-nav.js
```diff
@@ -6,16 +6,16 @@ const navigationSchema = require('./validationSchemas/navigation');
...
 const landingFile = path.resolve(__dirname, 'landing-navigation.json')
 const modulesFile = path.resolve(__dirname, 'fed-modules.json')
 
 
 async function validateLanding() {    
   try {
-    const file = fs.readFileSync(landingFile, {
-      encoding: 'utf-8'
-    })
+    const file = fs.readFileSync(landingFile, 'utf-8')
```
karelhala, line 17: This is completely unnecessary change. Both object and string works.

karelhala (changes requested):
We do want to have `rhel-navigation` instead of `insights-navigation` since there is planned change to switch to `rhel`.

chrome/validate-nav.js
redallen: Yes, but since I changed line 37 I decided to change them all to be consistent.

chrome/validate-nav.js
redallen: Done.

karelhala (approved):

PR #737, opened 2021-07-07
Revert "add cloud tutorials to nav, rename rhel-navigation to insights-navigation"

karelhala:
Reverts RedHatInsights/cloud-services-config#735


Hyperkid123:
@karelhala why was this reverted?

karelhala:
@Hyperkid123 we've added symlinks to insights and CI broke. I've reverted this to make the fix in assets puller easier to debug.

Hyperkid123:
@karelhala ah ok. It was not the symlink. The puller run out of memory and weird stuff started happening. It was addressed.

PR #740, opened 2021-07-08
Use new chrome schema (prod-beta)

fhlavac:
https://issues.redhat.com/browse/RHCLOUD-15094

Ported navigation to the new schema according to:
https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/docs/chrome/docs.md


Hyperkid123 (changes requested):
@fhlavac I have found small issues.
- Resource optimizations nav item is missing from insights bundle
- ROS app is not defined in modules

Hyperkid123 (approved):

PR #747, opened 2021-07-09
[ci-beta] Catalog module parameter

lgalis:


Hyperkid123:
@lgalis the `main.yml` file is no longer used for anything in chrome in ci/qa-beta environments. (others will follow in weekly interwals). You need to change the `fed-modules.json` file.

Here is an example of chrome 2 application: https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/docs/chrome/docs.md#registering-new-module-app-to-chrome

And here is the entry for catalog: https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/fed-modules.json#L387

PR #813, opened 2021-07-28
[prod-beta] Add ocp advisor module.

Hyperkid123:


gkarat:
thanks Martin! it's late, but LGTM 😸

PR #814, opened 2021-07-28
adding adding metadata to chrome modules for test app

Jason-RH:


Jason-RH:
@ryelo another PR for the onboarding test - forgot to update fed_modules

chrome/fed-modules.json
```diff
@@ -512,6 +512,18 @@
             }
         ]
     },
+    "onboarding": {
+        "manifestLocation": "/apps/test-onboarding-app/fed-mods.json",
```
Hyperkid123: From what I can tell the location should be 
```diff
-/apps/test-onboarding-app/fed-mods.json
+/apps/onboarding/fed-mods.json
```
It should follow the appId specified in `main.yml`

Jason-RH: Yeah I wasn't sure about that. I checked the fed-mods.json file in /dist and it contained:
` {
     "testOnboardingApp": {
         "entry": [
             "/beta/apps/test-onboarding-app/testOnboardingApp.ce69398eb950b7cbf563.js"
         ],
         "modules": [
             "/beta/apps/test-onboarding-app/js/412.85953f4f0a0c4e2bbf55.js"
         ]
     }
 }
`

Hyperkid123: Yeah, you want to add reference to that `fed-mods.json` file. The file is stored in CDN under the `appId` directory. And the appId matches the key in `main.yml`. 

So when I look at: https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/main.yml#L674 I see "onboarding".

That brings me to a question. When you look at the application `package.json`, what string do you see under `insights.appname` key? The key in `main.yml` should be the same. I think it might not match.

Jason-RH: Yeah they're different. I'll update

PR #822, opened 2021-07-29
Only sync netstorage

astrozzc:
Remove the old sync netstorage code since we no longer use this jenkins pipeline


ryelo:
@astrozzc We might just want to port this to all of the other branches too

ryelo (approved):

lphiri (commented):
@astrozzc  there is no going back after today - lets just delete the code instead of commenting it out.

PR #856, opened 2021-08-06
Adds malware nav entry

AllenBW:


chrome/rhel-navigation.json
```diff
@@ -145,6 +145,24 @@
...
                     ]
+                },
+                {
+                    "title": "Malware",
+                    "expandable": true,
+                    "routes": [
+                        {
+                            "appId": "malware",
+                            "title": "Signatures",
+                            "href": "/insights/malware/systems",
```
ryelo: are both of these hrefs supposed to be the same?

ryelo (changes requested):

chrome/rhel-navigation.json
AllenBW: nope

karelhala (approved):

PR #857, opened 2021-08-06
[CI Beta] Re-enable smoke tests

john-dupuy:
Tested against prod-beta/prod-stable with a "dummy" script here:
https://jenkins-jenkins.5a9f.insights-dev.openshiftapps.com/job/test-csc-deploy-pipeline/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -133,7 +133,7 @@ node {
             from: 'csc-jenkins@redhat.com',
             replyTo: 'rlong@redhat.com',
             subject: 'CSC build: Smoke tests failed',
-            to: 'rlong@redhat.com,khala@redhat.com,jdupuy@redhat.com'
+            to: 'csc-pipeline-failures@redhat.com'
       error('Smoke tests failed! Changes have been rolled back.')
     }
   }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,20 +5,20 @@ node {
   // Only run one build at a time; otherwise they'll fail since you can only have one Akamai activation at a time
   properties([disableConcurrentBuilds()])
 
-  stage ("create backup for old YAML files") {
+  stage ("Create Backup old YAML/JSON files") {
     BRANCH = env.BRANCH_NAME.replaceAll("origin/", "")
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -66,4 +64,67 @@ node {
...
+        }
+      }
+
+      // If smoke tests have errors
+      currentBuild.result = 'ABORTED'
+      mail body: "CSC smoke tests failed. Failed build is here: ${env.BUILD_URL}" ,
+            from: 'csc-jenkins@redhat.com',
+            replyTo: 'rlong@redhat.com',
+            subject: 'CSC build: Smoke tests failed',
+            to: 'rlong@redhat.com,khala@redhat.com,jdupuy@redhat.com'
```
karelhala: @john-dupuy what are your thoughts on creating a mailing list something like "CRC outages" and funneling all of these errors to this place?

ryelo:
@john-dupuy Curious here, is there a reason to create backup files anymore? I know we had to do it before because we would have to roll back akamai releases, but now that we're only syncing to netstorage, can we just fail the build without backing up files and just never actually push them to netstorage?

john-dupuy:
> @john-dupuy Curious here, is there a reason to create backup files anymore? I know we had to do it before because we would have to roll back akamai releases, but now that we're only syncing to netstorage, can we just fail the build without backing up files and just never actually push them to netstorage?

@ryelo In the example of some new app being added to CSC, I understand the pipeline looks like this:
1) App owner adds necessary info to main.yml and nav-json files through PR to e.g. prod-beta
2) PR gets merged
3) Netstorage gets synced with the new changes to nav-json and main.yml (at this point the new app will be live)

If this is correct, in order to test the change we have to sync netstorage first. Therefore we need to backup files and revert if the changes cause the test to fail.

Jenkinsfile
john-dupuy: I've opened a ticket with IT to create the mailing list and will change this message to use the list in a follow-up.

john-dupuy: Actually, seems the list has already been created, I'll make this change now.

karelhala (approved):

PR #874, opened 2021-08-25
[prod/beta] Show OCP Advisor application in the nav panel

gkarat:
Fixes https://issues.redhat.com/browse/CCXDEV-5017.

There were some updates to the app OCP Advisor and we now need to start showing OCP Advisor application in ci/beta and qa/beta environments.

### DIFF
--- a/chrome/openshift-navigation.json
+++ b/chrome/openshift-navigation.json
@@ -34,6 +34,24 @@
             "title": "Insights",
             "navItems": [
                 {
+                    "title": "Advisor",
+                    "expandable": true,
+                    "routes": [
+                        {
+                            "appId": "ocpAdvisor",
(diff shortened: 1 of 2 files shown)


chrome/openshift-navigation.json
```diff
@@ -33,6 +33,24 @@
             "groupId": "insights",
             "title": "Insights",
             "navItems": [
+                {
```
gkarat, line 36: This item was not added before for some reason (even when the app has been normally functioning before in prod/beta). I have copy-pasted this new object item from ci-beta version https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/openshift-navigation.json#L41.

Hyperkid123 (approved):

PR #875, opened 2021-08-25
add fedramp apps

ryelo:


Hyperkid123 (commented):
This will handle the nav, but the modules will be still accessible from URL. To prevent that we will need a flag in the `fed-modules.json` file.

And because we have the `appId` in the nav schemas, we can reference them when rendering the nav and hide items based on the flag in the modules specification. So I would probably use this flag in the `fed-modules.json` rather than the nav files.

Hyperkid123 (approved):

PR #927, opened 2021-10-04
update application settings;

ryelo:
https://issues.redhat.com/browse/RHCLOUD-15888

> A non-org admin user should be able to access whatever apps exist in the Applications section if they have the correct RBAC permissions. We should stop gating this nav item by org admin checks and only check for RBAC. 
> 
> Show Applications in the nav if user has permissions for Advisor and/or Cost Management.
> 
> If user doesn't have permissions for either, (ideally we will be able to) hide the entire Applications section in the nav.
> 
> There are currently only two applications in this section: 
> 
> Advisor - allow access if user has advisor admin (star:star)  permissions
> 
> Cost - allow access if user has cost-management admin (star:star) permissions

### DIFF
--- a/chrome/settings-navigation.json
+++ b/chrome/settings-navigation.json
@@ -78,15 +78,6 @@
         {
             "title": "Applications",
             "expandable": true,
-            "permissions": [
-                {
-                    "method": "isEntitled",
-                    "args": [
-                        "insights",
-                        "cost_management"
-                    ]
-                }
-            ],
             "routes": [
                 {
                     "appId": "applications",
--- a/chrome/settings-navigation.json
+++ b/chrome/settings-navigation.json
@@ -80,10 +80,10 @@
             "expandable": true,
             "permissions": [
                 {
-                    "method": "isOrgAdmin",
-                    "apps": [
-                        "cost-management",
-                        "insights"
+                    "method": "isEntitled",
+                    "args": [
+                        "insights",
+                        "cost-management"
                     ]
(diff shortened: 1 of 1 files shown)


Hyperkid123 (dismissed):

chrome/settings-navigation.json
```diff
@@ -82,7 +82,8 @@
                 {
                     "method": "isOrgAdmin",
```
Hyperkid123: There is still the org admin reference. Shouldn't this dondition be removed entirely?

karelhala: I think that this should be `isEntitled`

ryelo: Updated. Can you take another look? I want to make sure that this checks insights OR cost, not AND

chrome/settings-navigation.json
```diff
@@ -80,9 +80,10 @@
             "expandable": true,
             "permissions": [
                 {
-                    "method": "isOrgAdmin",
-                    "apps": [
-                        "cost-management"
+                    "method": "isEntitled",
+                    "args": [
+                        "insights",
```
Hyperkid123: @ryelo So the `isEentitled` takes only one parameter at one time, so this won't work. You can use multiple `isEntitled` methods, but that does `AND` so all permission must be `true` in order to show the item. So I guess we need `OR` evaluation then.

Hyperkid123 (approved):

PR #930, opened 2021-10-14
Add OpenShift Insights api v2

matysek:


main.yml
```diff
@@ -207,6 +207,7 @@ ccx:
   api:
     versions:
       - v1
+      - v2
```
karelhala, line 210: I've checked this in the app and looks like we support just one api version, the first one is listed in the table. So adding the `v2` before `v1` is the right way. I'll create a ticket in our JIRA to support multiple versions where we'd show multiple tags in the version column and users will be able to click on either of these values (the first one will be used as default when clicking on title in table)

tisnik (commented):
Seems ok

main.yml
matysek: @karelhala thanks. Our v2 api still evolves. So the default should be still the `v1` as it is in this MR. Thank you.

karelhala (approved):

PR #955, opened 2021-11-05
Add flat nav items of hac dynamic nav

karelhala:
merge after https://github.com/RedHatInsights/insights-chrome/pull/1685

### DIFF
--- a/chrome/hac-navigation.json
+++ b/chrome/hac-navigation.json
@@ -16,6 +16,10 @@
         "dynamicNav": "hacCore/dev-demo-section"
       },
       {
+        "title": "Flat nav",
+        "dynamicNav": "hacCore"
+      },
+      {
           "title": "App Studio",
           "appId": "hacDev",
           "href": "/hac/app-studio"


chrome/hac-navigation.json
```diff
@@ -15,6 +15,10 @@
         "title": "Demo Plugin",
         "dynamicNav": "hacCore/dev-demo-section"
       },
+      {
+        "title": "Flat nav",
+        "dynamicNav": "hacCore"
+      },
```
andrewballantyne, line 21: @karelhala Can you help me understand the need for this?

Also, did we want to remove the old navs or are they there as a migration path?

karelhala: @andrewballantyne the dynamic nav can be 2 parts `$appName` to get everything from the application, or in HAC case `$appName/section` this is to allow listing the different plugins in correct order. For example imagine we have 3 dynamic plugins
* `pluginA` has 1 section `mySection`
* `pluginB` has 2 sections `otherSection` and `anotherSection`
* `pluginC` has just flat navigation `myNav`

For instance if we want to have navigation like
* myNav
* mySection
* anotherSection
* otherSection

The config can be:
```
{
  title: 'Navigation from pluginC',
  dynamicNav: 'hacCore'
},
{
  title: 'Navigation from pluginB',
  dynamicNav: 'hacCore/mySection'
},
{
  title: 'Navigation from pluginB',
  dynamicNav: 'hacCore/anotherSection'
},
{
  title: 'Navigation from pluginB',
  dynamicNav: 'hacCore/otherSection'
},
```

Further we could also reference from which plugin the navigation should be pulled by adding `~`
```
{
  title: 'Navigation from pluginC',
  dynamicNav: 'hacCore/~pluginC'
},
```

This approach allows for easier fine tuning, we don't have to pull every manifest to build just the small part of navigation. Incremental navigation build vs big bang.

andrewballantyne: Does this information get converted into the Frontend CRs? Because this feels very static and not at all dynamic.

karelhala: Yup, this will be part of the CR for hac-core. This is just to tell chrome where to ask for dynamic navigation. The rest behind slash is arbitrary and just to tell us (HAC) what to load. We can have just one `dynamicNav` entry in here, but then we'd calculate the entire navigation from a few dynamic plugins. This would cause a bit of loading issue, while we'll calculate the navigation structure and everything. With record for smaller navigation chunks we can parallelize the loading and have the structure loaded a bit faster.

andrewballantyne (approved):

PR #960, opened 2021-11-11
Add akamai fast purge for config files.

Hyperkid123:
jira: https://issues.redhat.com/browse/RHCLOUD-16700

### Changes
- add cache purge to Jenkins 

cc @john-dupuy

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,7 @@ import groovy.json.JsonSlurper
 node {
   // Only run one build at a time; otherwise they'll fail since you can only have one Akamai activation at a time
   properties([disableConcurrentBuilds()])
-
+  NAVLIST = "ansible application-services docs insights landing openshift rhel settings user-preferences"
   stage ("Create Backup old YAML/JSON files") {
     BRANCH = env.BRANCH_NAME.replaceAll("origin/", "")
     if (BRANCH == "prod-stable") {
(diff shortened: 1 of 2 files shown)


karelhala (approved):
Codewise looks good!

Jenkinsfile
```diff
@@ -72,6 +72,17 @@ node {
           rsync -arv -e \"ssh -2\" ./chrome/*.json sshacs@cloud-unprotected.upload.akamai.com:${AKAMAI_APP_PATH}/chrome
         """
       }
+
+      // set akamai fast purge env
+      openShiftUtils.withJnlpNode(image: "quay.io/redhatqe/jenkins-slave-python36-akamai:v3.11") {
+        sh "wget https://raw.githubusercontent.com/RedHatInsights/cloud-services-config/ci-beta/akamai/cache_buster/bust_cache.py"
```
john-dupuy: Should we make this `env` aware? E.g. pulling from prod-beta/prod-stable instead of always ci-beta.

Jenkinsfile
```diff
@@ -72,6 +72,17 @@ node {
...
+        // get akamai fast purge credentials
+        withCredentials([file(credentialsId: "jenkins-eccu-cache-purge", variable: 'EDGERC')]) {
+          sh "python3 bust_cache.py $EDGERC ${BRANCH} ${NAVLIST}"
```
john-dupuy, line 79: If running on `stage-stable` or `stage-beta` - `NAVLIST` will be undefined.

Jenkinsfile
```diff
@@ -72,6 +72,17 @@ node {
           rsync -arv -e \"ssh -2\" ./chrome/*.json sshacs@cloud-unprotected.upload.akamai.com:${AKAMAI_APP_PATH}/chrome
         """
       }
+
+      // set akamai fast purge env
+      openShiftUtils.withJnlpNode(image: "quay.io/redhatqe/jenkins-slave-python36-akamai:v3.11") {
```
john-dupuy, line 75: So that it's clear in the pipeline stage view, I think it would be good to put this in a separate stage (if possible).

john-dupuy (commented):
Few comments, I tested running this Jenkinsfile (https://github.com/john-dupuy/cloud-services-config/blob/jdupuy-test-akamai-cache/Jenkinsfile) after fixing the `NAVLIST` and it ran fine. 

EDIT: Just thought of something - if the smoke tests fail and we revert the files, we need to purge the cache there too.

Jenkinsfile
Hyperkid123: We could. But I don't know if there will be any differences really.

john-dupuy (approved):
LGTM, thanks for this one @Hyperkid123!

Hyperkid123:
OK lets try this

PR #968, opened 2021-11-15
Add subscriptions->streams to application-services

zanewoodfin:
I'm attempting to add a "Subscriptions" dropdown to Application Services.  It should Have a small Insights heading above it and one entry in the dropdown for "Streams for Apache Kafka".  The wireframe is [here](https://marvelapp.com/prototype/5dhc6c0/screen/82172335) for reference.  Let me know if this looks like it should generate what's in the wireframe.

### DIFF
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -68,6 +68,24 @@
             ]
         },
         {
+            "groupId": "insights",
+            "title": "Insights",
+            "navItems": [
+                {
+                    "title": "Subscriptions",
+                    "expandable": true,
+                    "routes": [
+                        {
+                            "appId": "applicationServices",
+                            "title": "Streams for Apache Kafka",
+                            "isBeta": true,
+                            "href": "/application-services/subscriptions/streams"
+                        }
+                    ]
+                }
+            ]
+        },
+        {
             "appId": "applicationServices",
             "title": "Learning Resources",
             "href": "/application-services/learning-resources"
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -69,6 +69,7 @@
         },
         {
             "groupId": "insights",
+            "icon": "trend-up",
             "title": "Insights",
             "navItems": [
                 {
--- a/main.yml
+++ b/main.yml
@@ -763,8 +763,12 @@ application-services:
       - id: service-registry
         title: "Service Registry"
         group: overview
-      - id: streams
-        group: overview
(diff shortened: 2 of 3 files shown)


chrome/application-services-navigation.json
```diff
@@ -67,10 +67,28 @@
                 }
             ]
         },
+        {
+            "groupId": "insights",
```
Hyperkid123, line 71: You are missing the icon property. Based on the mocks, it should be
```js
"icon": "trend-up"
```

main.yml
```diff
@@ -763,8 +763,12 @@ application-services:
       - id: service-registry
         title: "Service Registry"
         group: overview
-      - id: streams
-        group: overview
+      - id: subscriptions
```
Hyperkid123: These changes are not required. Navigation is not using this config.

Hyperkid123 (commented):
This is how the nav currently looks.

screenshot-prod foo redhat com_1337-2021 11 16-09_55_53

cdcabrera:
Hey @Hyperkid123 is this file required to be updated with the app-services/subs/streams path?

https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/fed-modules.json#L291-L308

chrome/application-services-navigation.json
zanewoodfin: Added.

main.yml
zanewoodfin: Reverted.

When would I need to update this location?

Hyperkid123:
> Hey @Hyperkid123 is this file required to be updated with the app-services/subs/streams path?
> 
> https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/fed-modules.json#L291-L308

@cdcabrera sorry I was on PTO. Yes, you have to add a new entry to the modules in order to create a route in the applications services bundle. It can be the same shape as with the other two. But using the `/application-services/subscriptions` path.

Hyperkid123 (approved):

Hyperkid123:
@zanewoodfin when is this supposed to be merged?

zanewoodfin:
@Hyperkid123 I think this can be merged at any time because this is just going into ci-beta.

mirekdlugosz:
Looks good :+1:

RHOSAK is already available on qa-beta and stage-beta, so from Subscriptions point of view, we would be fine with promoting this commit to other beta branches.

Screenshot from 2021-11-19 15-34-45

PR #972, opened 2021-11-22
Changed the order of links in Insights Advisor Menu according to task

Fewwy:
Because of the RHEL/Insights Advisor unification, I need to change places these 2 menus.

### DIFF
--- a/chrome/openshift-navigation.json
+++ b/chrome/openshift-navigation.json
@@ -39,14 +39,14 @@
                     "routes": [
                         {
                             "appId": "ocpAdvisor",
-                            "title": "Clusters",
-                            "href": "/openshift/insights/advisor/clusters",
+                            "title": "Recommendations",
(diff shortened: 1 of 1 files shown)


Fewwy:
@karelhala Karel, please take a look at this patch. My changes will be enough to switch places 2 links in the menu or should I change something else?

karelhala (approved):
Yup, if we are interested just in changing the order, this should be just enough!

karelhala:
@Fewwy the PR has been merged and should be automatically be updated to `ci`, `qa`, `stage` and `qaprodauth` beta environments. Can you please check that everything as it should? Once you are happy with it it's generally a good thing to promote the PR to all other branches (usually the inital reporter creates PR from cherry-picking this code) - prod-beta, ci-stable, qa-stable and prod-stable.

PR #973, opened 2021-11-23
Ccxdev 6437 switch insights advisor menu order in prod beta

Fewwy:
The recommendations link should be first in the advisor-> insights menu like in RHEL.


Hyperkid123:
@Fewwy this won't work. Either cherry-pick the specific commit or just do the same changes. As you can see, there is a lot of differences between different environments. So we keep them separate.

Fewwy:
> @Fewwy this won't work. Either cherry-pick the specific commit or just do the same changes. As you can see, there is a lot of differences between different environments. So we keep them separate.

Yeah, let me close the PR and clear this. That was an accident

PR #982, opened 2021-11-29
image-builder: Drop permissions check

croissanne:
Image build has been in public beta since last week. The check can be dropped now.


Hyperkid123 (changes requested):
@Gundersanne  You will have to remove the check in the navigation file: https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/rhel-navigation.json#L64

The main.yml is no longer used for navigation rendering.

croissanne:
Done, can I drop it in both or should I not touch main.yml anymore?

Hyperkid123 (approved):

PR #998, opened 2021-12-14
[qa-stable] Automation Hub - drop My namespaces (#994)

himdel:
Problem:

https://console.stage.redhat.com/ansible/automation-hub still shows "My Namespaces" in left-side nav.

My original assumption was that stage-stable (#994) is the place to fix it, but.. maybe not?

`qa-stable` is the only branch with `My namespaces` now, so my guess is that either stage is deployed from qa-stable for some reason, or that something failed after merging #994 

@Hyperkid123 can you check please? :)

PR #999, opened 2021-12-16
Remove my namespaces from navigation config

karelhala:
### No need for my namespaces in navigation for ansible

Since we no longer need my namespaces in navigation, let's remove them.

ping @himdel

### DIFF
--- a/chrome/ansible-navigation.json
+++ b/chrome/ansible-navigation.json
@@ -26,12 +26,6 @@
                 },
                 {
                     "appId": "ansibleAutomationHub",
-                    "title": "My Namespaces",
-                    "href": "/ansible/automation-hub/my-namespaces",
-                    "product": "Ansible Automation Hub"
-                },
-                {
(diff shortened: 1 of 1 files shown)


himdel:
Ah, this certainly explains why #994 or #998 didn't do anything, thanks! :) :+1: 

Looks like `My Namespaces` are only present in `qa-stable` & `stage-stable`, looks like this should take care of both? :)

Hyperkid123 (approved):

himdel:
@Hyperkid123, @karelhala anything I can do to get this finished please? I'm still seeing this on stage-stable, https://github.com/RedHatInsights/cloud-services-config/blob/stage-stable/chrome/ansible-navigation.json#L27-L32

Hyperkid123:
@himdel should be OK now. It will take a few moments to sync on akamai.

himdel:
Thanks, that worked! :)

PR #1014, opened 2022-01-13
Add Baselines subapp for Patch application

mkholjuraev:
[SPM-1304](https://issues.redhat.com/browse/SPM-1304)

### DIFF
--- a/chrome/rhel-navigation.json
+++ b/chrome/rhel-navigation.json
@@ -149,6 +149,12 @@
                             "title": "Packages",
                             "href": "/insights/patch/packages",
                             "product": "Red Hat Insights"
+                        },
+                        {
+                            "appId": "patch",
+                            "title": "Patch set",
+                            "href": "/insights/patch/patch-set",
+                            "product": "Red Hat Insights"
                         }
                     ]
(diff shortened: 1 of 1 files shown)


main.yml
```diff
@@ -817,6 +817,8 @@ patch:
         title: Systems
       - id: packages
         title: Packages
+      - id: baselines
```
Hyperkid123: You have to update the navigation JSON. main.yml is no longer used for chrome navigation. You can follow this doc: https://github.com/RedHatInsights/cloud-services-config#chromebundle-navigationjson. The section also contains a link with a more detailed description.

Hyperkid123 (changes requested):

main.yml
mkholjuraev: I have mae some changes in the PR in the expected file. The navigation title was reworded to  'Patch set' by UX.

Hyperkid123 (approved):

Hyperkid123:
@mkholjuraev what environments is this change targetting? This PR will only affect ci/qa/stage beta envs. I can promote it further.

PR #1050, opened 2022-02-14
NOTIF-475 update menu to be consistent with bundle menu order in stage

aferd:


Hyperkid123 (commented):
@aferd can you please change the target branch to be `ci-beta`. Stage beta is just a mirror of ci-beta and merging the PR into stage beta directly could break the synchronization job.

PR #1052, opened 2022-02-15
 NOTIF-475 update menu to be consistent with bundle menu order in stage

aferd:


aferd:
If I would like to have this update in stage-stable as well, do I need to make another PR against ci-stable?

Hyperkid123:
> If I would like to have this update in stage-stable as well, do I need to make another PR against ci-stable?

I will take care of it.

Hyperkid123 (approved):

PR #1065, opened 2022-03-08
feat: added the acs config

sachaudh:
This PR is to add the Advanced Cluster Security (ACS) app to the consoledot environment

### DIFF
--- a/chrome/fed-modules.json
+++ b/chrome/fed-modules.json
@@ -623,15 +623,27 @@
         ]
     },
     "sandbox": {
-      "manifestLocation": "/apps/sandbox/fed-mods.json",
-      "modules": [
+        "manifestLocation": "/apps/sandbox/fed-mods.json",
+        "modules": [
+            {
+                "id": "sandbox",
+                "module": "./RootApp",
+                "routes": [
+                    "/openshift/sandbox"
+                ]
+            }
+      ]
+    },
+    "acs": {
+        "manifestLocation": "/apps/acs/fed-mods.json",
+        "modules": [
           {
-              "id": "sandbox",
+              "id": "acs",
               "module": "./RootApp",
               "routes": [
-                  "/openshift/sandbox"
+                  "/application-services/acs"
               ]
           }
       ]
-  }
+    }
 }
--- a/main.yml
+++ b/main.yml
@@ -11,6 +11,18 @@ access-requests:
     paths:
       - /internal/access-requests
 
+acs:
+  title: Advanced Cluster Security
+  api:
+    versions:
+      - v1
+  description: Frontend for Red Hat employees to manage ACS instances
+  deployment_repo: https://github.com/stackrox/acs-ui-build
+  frontend:
+    paths:
+      - /application-services/acs
+  source_repo: https://github.com/stackrox/acs-ui
+
 advisor:
   title: Advisor
   api:
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -68,6 +68,23 @@
             ]
         },
         {
+            "title": "Advanced Cluster Security",
+            "expandable": true,
+            "routes": [
+                {
+                    "appId": "applicationServices",
+                    "title": "ACS Instances",
+                    "isBeta": true,
+                    "href": "/application-services/acs"
+                },
+                {
+                    "title": "Documentation",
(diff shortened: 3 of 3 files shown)


main.yml
```diff
@@ -11,6 +11,18 @@ access-requests:
     paths:
       - /internal/access-requests
 
+acs:
+  title: Advanced Cluster Security
+  api:
+    version:
```
karelhala: A typo
```suggestion
    versions:
```

main.yml
```diff
@@ -11,6 +11,18 @@ access-requests:
...
+      - v1
+  description: Frontend for Red Hat employees to manage ACS instances
+  deployment_repo: https://github.com/stackrox/acs-ui-build
```
karelhala, line 20: @ryelo Are there any drwbacks that the build repo is outside our org? AFAIK it shouldn't be a problem, just maintenance reasons.

main.yml
```diff
@@ -11,6 +11,18 @@ access-requests:
...
+  frontend:
+    paths:
+      - /acs
```
karelhala: This should include the full path
```suggestion
      - /application-services/acs
```

chrome/fed-modules.json
```diff
@@ -623,15 +623,27 @@
...
+    "acs": {
+        "manifestLocation": "/apps/acs/fed-mods.json",
+        "modules": [
           {
-              "id": "sandbox",
+              "id": "acs",
               "module": "./RootApp",
               "routes": [
-                  "/openshift/sandbox"
+                  "/acs"
```
karelhala: This should include the full path
```suggestion
                  "/application-services/acs"
```

karelhala (commented):
Looking good! Just a few routing adjustments and one typo.

main.yml
Hyperkid123: I am helping @sachaudh with the build repo. We will see if Jenkins picks up the changes. If not we will discuss an acceptable solution.

sachaudh: If it's better to keep it within the org, would there be any trouble in moving the repo after adding this? If so, I can try doing that before

karelhala: I am just worried for jenkins to pick up the changes. But if jenkins will be fine we def can keep it where it is right now (no need to stall this any further)

karelhala:
I don't have any objections to the code or anything, @sachaudh do you want us to merge this so you can test if everything is running?

sachaudh:
> I don't have any objections to the code or anything, @sachaudh do you want us to merge this so you can test if everything is running?

Yes, that would be very much appreciated 🙏🏼

main.yml
ryelo: AFAIK the build repo has to be within the RedHatInsights github org. There shouldn't be any trouble moving the repo if it doesn't pick it up.

Hyperkid123:
@sachaudh lets see if the Jenkins pipeline picks up the changes

main.yml
sachaudh: No worries. I'm not an organization member so what should I do to create a repo under RedHatInsights?

sachaudh:
> @sachaudh lets see if the Jenkins pipeline picks up the changes

@Hyperkid123 how can I test to see if it worked. I wasn't able to load It locally before because I didn't configure it in the cloud-services-config. Now that it's in, I noticed that the side nav bar still didn't have the link and it redirects me saying that path doesn't exist

Screen Shot 2022-03-14 at 10 04 59 AM

sachaudh:
Oops my bad, so it does work actually. I was supposed to check `ci.cloud.redhat.com/beta`. Nice 👍🏼 

Screen Shot 2022-03-14 at 10 10 27 AM

But yea, I'm wondering how can I get it working locally with the new appUrl?

Hyperkid123:
@sachaudh 

> Oops my bad, so it does work actually. I was supposed to check ci.cloud.redhat.com/beta. Nice 👍🏼

You should not use ci/qa environments. If you can use `console.stage.redhat.com/beta`. It shares the config with ci and qa

> But yea, I'm wondering how can I get it working locally with the new appUrl?

That depends on your local configuration.
https://github.com/stackrox/acs-ui/blob/main/config/dev.webpack.config.js The config looks OK. I think you just have to run the app in the BETA env.  You want to start the dev server with the `npm run:start:beta` command.

PR #1066, opened 2022-03-09
Add Vulnerability for OpenShift as a sub-app of OpenShift

leSamo:


Hyperkid123 (commented):
This change will have no effect. Navigation is configured through the JSON files in the `chrome` directory. You can follow this guide: https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/docs/chrome/docs.md#adding-a-new-top-level-link-to-the-chrome-left-navigation

PR #1076, opened 2022-03-30
Update Application settings

aneelac22:
[Jira ref RHCLOUD-15888]


Hyperkid123 (changes requested):
@aneelac22 can you please change the target branch to `ci-beta`. It will get synced to the correct branches.

fhlavac (approved):

Hyperkid123 (approved):

PR #1088, opened 2022-04-13
add app-studio-docs to fed-modules

InsaneZein:


Hyperkid123 (commented):
@InsaneZein This looks very good. We now have to add a navigation entry to the [hac-navigation](https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/hac-navigation.json) file. We can merge after that.

chrome/hac-navigation.json
```diff
@@ -23,6 +23,11 @@
           "title": "App Studio",
           "appId": "hacDev",
           "href": "/hac/app-studio"
+      },
+      {
+        "title": "App Studio Docs",
+        "appId": "app-studio-docs",
```
Hyperkid123: The `appId` does not match the one in the modules definition `appStudioDocs`

Hyperkid123 (approved):

PR #1098, opened 2022-04-20
Add image-builder to stage-stable

croissanne:
There will be a followup where it's added to the nav as well within the next week or so, our GA date is targeted 2nd of May.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - node
+  - 16
 before_install:
   - npm install -g npm@latest
 install:
--- a/main.yml
+++ b/main.yml
@@ -450,6 +450,23 @@ frontend-assets:
   deployment_repo: https://github.com/RedHatInsights/frontend-assets-build.git
   source_repo: https://github.com/RedHatInsights/frontend-assets
 
+image-builder:
+  title: Image Builder
+  api:
+    versions:
+      - v1
+  deployment_repo: https://github.com/RedHatInsights/image-builder-frontend-build
+  frontend:
+    section: operations
+    module:
+      appName: image-builder
+      scope: image_builder
+      module: ./RootApp
+    title: Image Builder
+    paths:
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - node
+  - 16
```
karelhala, line 3: any reason for this change?

main.yml
```diff
@@ -450,6 +450,23 @@ frontend-assets:
   deployment_repo: https://github.com/RedHatInsights/frontend-assets-build.git
   source_repo: https://github.com/RedHatInsights/frontend-assets
 
+image-builder:
+  title: Image Builder
+  api:
+    versions:
+      - v1
+  deployment_repo: https://github.com/RedHatInsights/image-builder-frontend-build
+  frontend:
```
karelhala, line 459: No need for this frontend config anymore, can you please however update the rhel navigation to include the image builder? https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/rhel-navigation.json#L60-L63 you will also have to add https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/fed-modules.json#L593-L604

karelhala (commented):
Thank you for the PR, the `stage-stable` branch is synched from `qa-stable`. Can you please close this PR and re-open new one with base branch as `qa-stable`. Thank you!

.travis.yml
Hyperkid123: yes

The installed version is borked. I have noticed it in another PR.

main.yml
croissanne: Since our GA date is may 2nd I thought to postpone the nav change a little bit longer. However for stage we can already enable it? And then production next week or so?

karelhala: Yup, we can promote this to stage-stable. You can iterate on it and once you have GA we'll merge the prod-stable PR

PR #1099, opened 2022-04-20
Update image-builder declaration

croissanne:
There will be a followup where it's added to the nav as well within the next week or so, our GA date is targeted 2nd of May.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -447,18 +447,20 @@ inventory:
 
 image-builder:
   title: Image Builder
+  api:
+    versions:
+      - v1
   deployment_repo: https://github.com/RedHatInsights/image-builder-frontend-build
-  permissions:
-    method: apiRequest
-    args:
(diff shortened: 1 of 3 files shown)


main.yml
```diff
@@ -447,18 +447,20 @@ inventory:
 
 image-builder:
   title: Image Builder
+  api:
+    versions:
+      - v1
   deployment_repo: https://github.com/RedHatInsights/image-builder-frontend-build
-  permissions:
```
Hyperkid123, line 451: The permissions must be removed from the navigation files as well. This one is not used

croissanne: image-builder isn't in the nav yet for prod-stable. I wanted to wait with that since our GA date is may 2nd (thought of doing it like 28th of april).

Hyperkid123 (approved):

PR #1105, opened 2022-04-21
Add image-builder federated module

croissanne:
Oversight from https://github.com/RedHatInsights/cloud-services-config/pull/1100, as I think this is strictly required? 

Also noticed we dont' follow the `-` => camelcase convention, but I think we'll need to duplicate our own federated module config for that and expose it both as image_builder and imageBuilder.

### DIFF
--- a/chrome/fed-modules.json
+++ b/chrome/fed-modules.json
@@ -512,5 +512,17 @@
                 ]
             }
         ]
+    },
+    "image_builder": {
+        "manifestLocation": "/apps/image-builder/fed-mods.json",
+        "modules": [
+            {
+                "id": "image-builder",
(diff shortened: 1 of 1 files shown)


Hyperkid123:
> Also noticed we dont' follow the - => camelcase convention, but I think we'll need to duplicate our own federated module config for that and expose it both as image_builder and imageBuilder.

@Gundersanne that is actually ok. The issue is with the dash character `-` not underscore `_`. Webpack should be able to use is just fine.

Hyperkid123 (approved):

croissanne:
Yea but our insights.appname is still "image-builder". I guess using camelcase instead of `_` is just a convention though?

PR #1107, opened 2022-04-21
Adding nav item for insights settings.

arivepr:
Self titled! This is addressing part of [this ticket.](https://issues.redhat.com/browse/RHCLOUD-18230)

### DIFF
--- a/chrome/settings-navigation.json
+++ b/chrome/settings-navigation.json
@@ -151,6 +151,27 @@
                             ]
                         }
                     ]
+                },
+                {
+                    "appId": "applications",
+                    "title": "Insights",
+                    "href": "/settings/applications/insights",
+                    "permissions": [
+                        {
+                            "method": "isEntitled",
+                            "args": [
+                                "insights"
+                            ]
+                        },
+                        {
+                            "method": "hasPermissions",
+                            "args": [
+                                [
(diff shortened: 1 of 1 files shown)


chrome/settings-navigation.json
```diff
@@ -151,6 +151,27 @@
...
+                    "href": "/settings/applications/insights",
+                    "permissions": [
+                        {
+                            "method": "isEntitled",
+                            "args": [
+                                "insights"
+                            ]
+                        },
+                        {
+                            "method": "hasPermissions",
```
Hyperkid123, line 167: Do we need both entitlements and permissions check? Shouldn't be entitlements enough? or do we need a user that actually can do everything in the insights space? Then probably just the permissions check should be enough

arivepr: I'm going to be honest, I have no idea haha. 
I was following the pattern that I saw with the rest of the settings page.

Though this PR is irrelevant for the task, I was told the settings/insights page is a nono, it just needs to redirect to settings/advisor, which is already defined in this file 😄

Hyperkid123: I believe that if a user has `"insights:*:*"` permissions they also must be entitled. @ryelo can you confirm that?

PR #1117, opened 2022-04-27
Add filterable edge inventory

karelhala:
Screenshot from 2022-04-27 16-23-22

### DIFF
--- a/chrome/edge-navigation.json
+++ b/chrome/edge-navigation.json
@@ -4,7 +4,6 @@
   "navItems": [
     {
       "title": "Inventory",
-      "filterable": false,
       "expandable": true,
       "routes": [
         {


chrome/edge-navigation.json
```diff
@@ -4,7 +4,7 @@
   "navItems": [
     {
       "title": "Inventory",
-      "filterable": false,
+      "filterable": true,
```
Hyperkid123: You can remove the flag. It's true by default. I don't want to make an impression that people have to include the flag in order to appear in the dropdown.

Hyperkid123 (approved):

chrome/edge-navigation.json
karelhala: Good point!

PR #1118, opened 2022-04-28
[prod-beta] Remove catalog and approval

lgalis:


Hyperkid123 (commented):
@lgalis is the catalog and approval supposed to be removed from the navigation as well? If so can you remove it from the JSON files in the chrome directory?

Hyperkid123 (approved):
@lgalis when do you want this change to go live?

lgalis:
@Hyperkid123 - it can be merged anytime to prod-beta. I will create one for stable, that can be merged anytime as well. Then prod-stable should be merged May 1st.

PR #1125, opened 2022-04-28
[summit] Add image-builder to RHEL navigation

croissanne:


Hyperkid123 (approved):
@Gundersanne when would you like us to merge the change? Is this a summit release or can we deploy the changes now?

croissanne:
> @Gundersanne when would you like us to merge the change? Is this a summit release or can we deploy the changes now?

This is a summit release yea, happy to do what the rest of insights is doing here. My understanding was that with the dev freeze  we need to get this in by may 4th.

My idea was to make sure everything is in place before the freeze/summit so we can deal with any hiccups if needed.

Hyperkid123:
Cool we will merge it on the May 4th the as we do with the rest of prod-stable changes

Hyperkid123 (approved):

PR #1137, opened 2022-05-02
Fix API docs for sources

lpichler:
This change adds open api docs for sources to this page https://console.stage.redhat.com/docs/api .

Our openapi specs is here https://console.stage.redhat.com/api/sources/v3.1/openapi.json 

Thanks @Hyperkid123!

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -1012,7 +1012,7 @@ sources:
   title: Discovered Inventory
   api:
     versions:
-      - v1
+      - v3.1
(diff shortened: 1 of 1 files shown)


karelhala (approved):
Looking good! Just FYI this will only add it to stage-beta. You should also open PRs to prod-beta, ci-stable and prod-stable as well.

Hyperkid123 (approved):

PR #1160, opened 2022-05-03
RHIF-38 [CI-STABLE] AAP Navbar changes for summit

adonispuente:
In accordance to RHIF-38, it should look like this with the addition of policies
Screen Shot 2022-05-03 at 2 16 40 PM

### DIFF
--- a/chrome/ansible-navigation.json
+++ b/chrome/ansible-navigation.json
@@ -39,76 +39,9 @@
             ]
         },
         {
-            "groupId": "operations",
-            "icon": "wrench",
-            "title": "Operations Insights",
-            "navItems": [
-                {
-                    "title": "Advisor",
-                    "expandable": true,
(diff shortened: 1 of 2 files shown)


chrome/ansible-navigation.json
```diff
@@ -154,36 +87,84 @@
...
+                        {
+                            "appId": "drift",
+                            "title": "Baselines",
+                            "href": "/ansible/drift/baselines",
+                            "product": "Red Hat Insights"
+                        }
+                    ]
+                },
+                {
+                    "appId": "policies",
```
Hyperkid123, line 149: This item is not on the preview

jiridostal: https://issues.redhat.com/browse/RHIF-38 - it's been discussed and Policies will be put back in, this change is correct.

PR #1181, opened 2022-05-12
[CCXDEV-8023] Add v2 version to the aggregator

juandspy:
We have a second version of the API at https://console.redhat.com/api/insights-results-aggregator/v2/openapi.json. Let's start showing it too.

PR #1185, opened 2022-05-17
Add v2 version to the aggregator to prod-stable

juandspy:
Add https://github.com/RedHatInsights/cloud-services-config/pull/1181 to prod-stable.

### DIFF
--- a/main.yml
+++ b/main.yml
@@ -170,6 +170,7 @@ ccx:
         title: Insights Results Aggregator
         versions:
           - v1
+          - v2
       gathering:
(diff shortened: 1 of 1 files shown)


juandspy:
The new API version is already visible in stage:
image
We can continue with prod.

Hyperkid123 (approved):

juandspy:
@Hyperkid123 or @karelhala , it looks like prod doesn't display the v2 API. Seems like you have to cherry pick the commit that added this in prod-stable :)

PR #1188, opened 2022-05-17
Side nav changes

h-kataria:
- changed urls from ansible/insights/ to ansible/automation-analytics

https://issues.redhat.com/browse/AA-1114

### DIFF
--- a/chrome/fed-modules.json
+++ b/chrome/fed-modules.json
@@ -379,6 +379,7 @@
                 "id": "ansible-automation-analytics",
                 "module": "./RootApp",
                 "routes": [
+                    "/ansible/automation-analytics",
                     "/ansible/insights",
                     "/ansible/reports",
                     "/ansible/savings-planner",
--- a/chrome/ansible-navigation.json
+++ b/chrome/ansible-navigation.json
@@ -45,43 +45,43 @@
                 {
                     "appId": "automationAnalytics",
                     "title": "Reports",
-                    "href": "/ansible/insights/reports",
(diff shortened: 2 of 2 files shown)


Hyperkid123 (dismissed):
The `automationAnalytics` module does not have route registered for `/ansible/automation-analytics`

Can you please update the `fed-modules.json` file?

karelhala (commented):
You should also update `automationAnalytics` federated module in `fed-mods.json` with `/ansible/automation-analytics` route. Otherwise this will throw an error of unknown application. Can we also test this on `prod-beta` first? A different PR to prod-beta would be required so we can test if everything works and perhaps could roll back if it would throw some errors.

chrome/fed-modules.json
```diff
@@ -379,7 +379,7 @@
                 "id": "ansible-automation-analytics",
                 "module": "./RootApp",
                 "routes": [
-                    "/ansible/insights",
+                    "/ansible/automation-analytics",
                     "/ansible/reports",
```
Hyperkid123, line 384: What about the rest of the routes on the list? Is the automation analytics app supposed to render there? I would guess no based on the links from the nav.

h-kataria: @Hyperkid123 not sure how are they working currently in that case, I was confused hence did not change rest of them

Hyperkid123: OK lest take clean them up in another PR

karelhala (approved):

PR #1191, opened 2022-05-18
Updated route

h-kataria:
https://issues.redhat.com/browse/AA-1114

@karelhala @Hyperkid123

### DIFF
--- a/chrome/fed-modules.json
+++ b/chrome/fed-modules.json
@@ -392,6 +392,7 @@
                 "id": "ansible-automation-analytics",
                 "module": "./RootApp",
                 "routes": [
+                    "/ansible/automation-analytics",
                     "/ansible/insights",
                     "/ansible/reports",
                     "/ansible/savings-planner",


chrome/fed-modules.json
```diff
@@ -392,7 +392,7 @@
                 "id": "ansible-automation-analytics",
                 "module": "./RootApp",
                 "routes": [
-                    "/ansible/insights",
+                    "/ansible/automation-analytics",
```
karelhala, line 395: No need to remove the `insights` route (it would actually cause errors in current setup)
```suggestion
                    "/ansible/automation-analytics",
                    "/ansible/insights",
```

h-kataria:
@karelhala made requested change, do i need to change [these](https://github.com/h-kataria/cloud-services-config/blob/updated_route_on_prod-beta/chrome/fed-modules.json#L397-L403) as well ?

karelhala:
@h-kataria well that's up to you. Is the app available on these URLs? If not, we can remove it by all means.

h-kataria:
@jlmitch5 do you have any advice on https://github.com/RedHatInsights/cloud-services-config/pull/1191#issuecomment-1131674681

https://github.com/RedHatInsights/cloud-services-config/pull/1188#discussion_r876676637

karelhala (approved):

PR #1193, opened 2022-05-19
Removing Edge

theute:


Hyperkid123 (commented):
@theute The code looks good. Do we have a ticket or can we get an ACK from someone on the edge team as confirmation they know this is going on?

theute:
Not sure we have a ticket, but this is not going to prod, this is why we are removing from stage

Hyperkid123:
OK then.

PR #1209, opened 2022-06-13
Add patch v2 api

psegedy:


karelhala (approved):
You should also open PRs for prod-beta, ci-stable and prod-stable to reflect this change in all envs.

psegedy:
sure, I'll do that but only for ci-stable for now since we don't have v2 deployed in prod

PR #1218, opened 2022-06-15
Adding the API Designer navigation item

EricWittmann:
We are adding a new application services managed service called API Designer.  We need a navigation menu on the left for it.  Here is the JIRA with approvals:

https://issues.redhat.com/browse/RHCLOUD-19795

I think we may still be working on exactly what the terminology will be for the sub-menu item.  I've take a stab at it for now and we can change it if needed (currently I've made it `API and Schema Designs` to try and fit in with the existing pattern as best I can).

### DIFF
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -19,6 +19,22 @@
       "icon": "cloud",
       "navItems": [
         {
+          "title": "API Designer",
+          "expandable": true,
+          "routes": [
+            {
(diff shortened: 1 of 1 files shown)


Hyperkid123 (approved):

tiffanynolan:
This also needs to have the   "isBeta": true, value set.

Hyperkid123:
@tiffanynolan this nav is only for the beta environment. Dunno if that changes things

PR #1231, opened 2022-06-28
feat: adding acs to prod-beta

sachaudh:
This PR is to add the Advanced Cluster Security (ACS) app to the consoledot (`prod-beta`) environment

### DIFF
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -102,6 +102,23 @@
       ]
     },
     {
+      "title": "Advanced Cluster Security",
+      "expandable": true,
+      "routes": [
+        {
+          "appId": "applicationServices",
+          "title": "ACS Instances",
+          "isBeta": true,
+          "href": "/application-services/acs"
+        },
+        {
+          "title": "Documentation",
+          "isExternal": true,
+          "href": "https://docs.openshift.com/acs/3.66/welcome/index.html"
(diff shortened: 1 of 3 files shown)


sachaudh:
So I wanted to just confirm that this is ok. Our backend team has created a public Fleet Manager API that apparently uses sso.redhat.com to authenticate the user. Since running the development environment locally uses a staging environment, the  UI uses sso.stage.redhat.com to authenticate the user. The token would therefore not work with API calls. They suggested we use sso.redhat.com instead so I thought I could run the UI locally to use sso.redhat.com instead. I got that part to work, but noticed that I'd need to make this change to actually see the ACS page in the UI too

I also requested they try to use sso.stage.redhat.com so I'm waiting to see if it's possible. In the meantime, I wanted to try it this way just so I had a Plan B in case they could not

karelhala:
Just FYI once this PR is merged the app will be visible on console.redhat.com/beta I would advise using the stage SSO and if not possible this as a fallback. We could also hide the navigation parts by some rules.

Hyperkid123 (approved):
Changes are good. @sachaudh can you please answer @karelhala question before we proceed with the deployment?

sachaudh:
Looks like we can get it working on staging now, so we can close this pull request. No need to merge. Thanks! 🙏🏼

PR #1244, opened 2022-07-26
rename patch navigation "Patch set"

mkholjuraev:
UX requested to rename 'Patch set' nav to 'Templates'


Hyperkid123 (approved):

mkholjuraev:
@Hyperkid123 thank you for the very quick response&review.

PR #1260, opened 2022-08-09
Add Cloud Connect

jblashka:
Adding Cloud Connect. This is the react app responsible for talking to the IT backend during the Cloud Marketplace customer purchase onboarding process

### DIFF
--- a/chrome/settings-navigation.json
+++ b/chrome/settings-navigation.json
@@ -106,27 +106,6 @@
             "routes": [
                 {
                     "appId": "applications",
-                    "title": "Advisor",
-                    "href": "/settings/applications/advisor",
-                    "permissions": [
-                        {
-                            "method": "isEntitled",
-                            "args": [
-                                "insights"
-                            ]
-                        },
-                        {
-                            "method": "hasPermissions",
-                            "args": [
-                                [
-                                    "advisor:*:*"
-                                ]
-                            ]
-                        }
-                    ]
-                },
-                {
-                    "appId": "applications",
                     "title": "Cost Management",
                     "href": "/settings/applications/cost-management",
                     "permissions": [


chrome/fed-modules.json
```diff
@@ -699,5 +699,17 @@
               ]
           }
       ]
+    },
+    "connect": {
+        "manifestLocation": "/apps/connect/fed-mods.json",
```
Hyperkid123: The manifest location depends on your `insights.appname` from your `package.json`

So it should be 

```
"manifestLocation": "/apps/statosphere-plf/fed-mods.json",
```

karelhala (commented):
This is effectively a new bundle, isn't it? You'll need a new navigation file as well called `connect-navigation.json`.

```JSON
{
    "id": "connect",
    "title": "Cloud connect",
    "navItems": [
        {
            "appId": "connect",
            "title": "Connect",
            "filterable": false,
            "href": "/connect"
        },
   ]
}
```

jblashka:
This is a bit of an odd one because there's intentionally no way to navigate to this app. It's only used during the purchase flow from AWS & Azure; if a user isn't coming directly from those Cloud Marketplaces via a special link generated within the Cloud Marketplaces, we can't do anything.

chrome/fed-modules.json
```diff
@@ -699,5 +699,17 @@
               ]
           }
       ]
+    },
+    "connect": {
```
Hyperkid123: The key should be `statospherePlf`. Again it depends on the appname, but due to some browser env limitations we can't support the dash character `-`. That is why its converted to camelCase

Hyperkid123:
> This is effectively a new bundle, isn't it? You'll need a new navigation file as well called `connect-navigation.json`.
> 
> ```json
> {
>     "id": "connect",
>     "title": "Cloud connect",
>     "navItems": [
>         {
>             "appId": "connect",
>             "title": "Connect",
>             "filterable": false,
>             "href": "/connect"
>         },
>    ]
> }
> ```

@karelhala this bundle will not have left navigation. It will also need a bit more different layout.

chrome/fed-modules.json
jblashka: I was going to update that insights.appname to connect as well

Hyperkid123: Ah, then it's OK. And you can use the `connect` keys/path

Hyperkid123 (approved):

PR #1293, opened 2022-09-23
feat: adding acs overview page

sachaudh:
The following PR will add another nav item for linking to our ACS overview page

### DIFF
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -124,9 +124,14 @@
       "routes": [
           {
               "appId": "acs",
+              "title": "Overview",
+              "href": "/application-services/acs/overview"
+          },
+          {
+              "appId": "acs",
               "title": "ACS Instances",
               "isBeta": true,
-              "href": "/application-services/acs"
+              "href": "/application-services/acs/instances"
           },
           {
               "title": "Documentation",


chrome/application-services-navigation.json
```diff
@@ -122,11 +122,16 @@
...
+              "appId": "acs",
+              "title": "Overview",
+              "href": "/application-services/acs/overview"
+          },
           {
               "appId": "acs",
               "title": "ACS Instances",
               "isBeta": true,
-              "href": "/application-services/acs"
+              "href": "/application-services/acs/instances"
```
vjwilson, line 134: Question for the folks who have been doing console-dot longer than ACS:

I noticed that the Kafka plugin doesn't have a "root" page, but if you manually enter
https://console.redhat.com/beta/application-services/streams
it redirects to 
https://console.redhat.com/beta/application-services/streams/kafkas

I don't see anything in Kafka's section of this file that shows how they did that.

What does ACS need to do to have similar redirect behavior?

sachaudh: @vjwilson I think we can do the redirection on our end in this file https://github.com/RedHatInsights/acs-ui/blob/main/src/Routes.js. I guess Kafka redirects the root path to the instances page and shows a 404 for non-matching routes. We can do that too. Right now, it'll redirect any non-matching route to the instances page, so it'll redirect the root path to the instances page

sachaudh:
@ryelo @Hyperkid123 @karelhala we're hoping to get this PR and the one on `prod-beta` in by Monday so we can see if everything works before our service preview on Tuesday. Let us know what you'd need from us to get this in. Thanks 🙏🏼

ryelo (approved):

PR #1297, opened 2022-09-27
Expose Provisioning wizard

ezr-ondrej:
My goal here: Another app being able to open up modal, that my app defines as a module.

I guess it's bit specific, so I'm not sure how to achieve this.
Ideally my app is not in the menu at all and is only opened from other app.

What should I do, is this correct? :)

Probably depends on https://github.com/RedHatInsights/frontend-components/pull/1623 & https://github.com/RHEnVision/provisioning-frontend/pull/118

### DIFF
--- a/chrome/fed-modules.json
+++ b/chrome/fed-modules.json
@@ -302,6 +302,10 @@
             }
         ]
     },
+    "provisioning": {
+        "manifestLocation": "/apps/provisioning/fed-mods.json",
+        "modules": []
+    },
     "subscriptions": {
         "manifestLocation": "/apps/subscriptions/fed-mods.json",
         "modules": [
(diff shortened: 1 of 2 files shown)


karelhala (approved):
@ezr-ondrej yes, this is correct. The main.yml update will host the static asset files, whereas the fed-modules.json will allow any app to import the launch-wizard app trough scalprum loader.

chrome/fed-modules.json
```diff
@@ -302,6 +302,15 @@
             }
         ]
     },
+    "provisioning": {
+        "manifestLocation": "/apps/provisioning/fed-mods.json",
+        "modules": [
```
Hyperkid123: I think even the modules array can be empty if we don't need any routes. The module is still going to be accessible vial the scalprum loader. This is just for chrome to know about the remote container. But it does not have to prepare any routes.

ezr-ondrej: right! make sense, removed, thanks ! :)

PR #1298, opened 2022-09-28
Add Malware detection app to prod-stable with isBeta flag

leSamo:
[ESSNTL-3360](https://issues.redhat.com/browse/ESSNTL-3360)


Hyperkid123 (changes requested):
@leSamo there is already a `malware` key in the main.yml

Hyperkid123 (approved):

PR #1299, opened 2022-09-29
[stage] Remove Edge bundle from User preferences

josejulio:


Hyperkid123 (approved):

josejulio:
Is `ci-stable` still going to `stage-stable` ? 
I can still see the `Edge` menu in https://console.stage.redhat.com/user-preferences/notifications/application-services

Hyperkid123:
Yes, it does. But may take a while.

PR #1309, opened 2022-10-13
first attempt at user-access redirects

zwitter1:


chrome/fed-modules.json
```diff
@@ -51,7 +51,7 @@
                 "id": "user-access",
                 "module": "./RootApp",
                 "routes": [
-                    "/settings/rbac"
+                    "/iam/user-access"
```
Hyperkid123, line 54: Can we keep both routes for now? This change will mess up all QE tests that are currently in place.

PR #1311, opened 2022-10-18
Use just one record for app services API

karelhala:


karelhala:
ping @wtrocki and @riccardo-forina to confirm if this is correct.

wtrocki:
Seems fine

riccardo-forina:
I think so? 👀 I suppose this can be tested on stage before going live right?

Hyperkid123:
@riccardo-forina yup this will land in stage beta first

Hyperkid123 (approved):

PR #1321, opened 2022-11-02
[stage-beta] Add Subscription Watch OpenShift Data Science

cdcabrera:
Activates OpenShift Data Science for Subs Watch
[SWATCH-167]

@Hyperkid123 et all, If the `appId` needs to be updated let me know.

It's escaping our memory on why the `appId` for these is `applicationServices` when the other configs for Subscriptions under `insights/rhel` and `openshift`  reflect the `subscriptions` `appId` ... we don't use the `applicationServices` appId anywhere within our GUI

### DIFF
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -150,9 +150,14 @@
           "expandable": true,
           "routes": [
             {
-              "appId": "applicationServices",
+              "appId": "subscriptions",
               "title": "Streams for Apache Kafka",
               "href": "/application-services/subscriptions/streams"
+            },
+            {
+              "appId": "subscriptions",
+              "title": "OpenShift Data Science",
+              "href": "/application-services/subscriptions/rhods"
             }
           ]
         }


Hyperkid123:
@cdcabrera as long as the application services shell knows about the route it should be ok. But if this is an actual subs app, it has to use that appId

cdcabrera:
@Hyperkid123 So both of those aspects are product views from Subs Watch, I'll go ahead and flip them over to `subscriptions`, much appreciated!

Hyperkid123:
@cdcabrera if that is the case you also need a correct route for the subscriptions module in the `fed-modules.json` so the chrome creates new route for the module.

cdcabrera:
@Hyperkid123 I might need some help with that, I see we already have a reference regarding `application-services` which also aligns for `insights` and `openshift` so I'm not super clear what else needs to be added beyond that route... if you could expand on what route?

https://github.com/RedHatInsights/cloud-services-config/blob/ci-beta/chrome/fed-modules.json#L323-L343
```
       ...{
              "id": "application-services-subscriptions",
              "module": "./RootApp",
              "routes": [
                  "/application-services/subscriptions"
              ]
          },
```

I do see someone added a `streams` reference but ours falls under `/application-services/subscriptions/streams`... so not even sure that's an aliased route, so we're assuming that's another app. In hindsight using generic naming was probably not such a great idea for us
```
"applicationServices": {
        "manifestLocation": "/apps/application-services/fed-mods.json",
        "modules": [
            {
                "id": "application-services",
                "module": "./RootApp",
                "routes": [
                    "/application-services",
                    "/application-services/streams"
                ]
            }
        ]
    },
```

Hyperkid123:
@cdcabrera Ah I see. Yeah, you are covered then.

Hyperkid123 (approved):

PR #1322, opened 2022-11-07
Fix reference to content-sources in main.yaml

jlsherrill:


jlsherrill:
We noticed that the apidoc link here:  https://console.stage.redhat.com/beta/docs/api/  is wrong, and this seemed incorrect, but i'm not sure if these two are related :)

jlsherrill:
ahh yes, based on https://github.com/RedHatInsights/api-frontend#api-docs  this does looks like the right place

and you can see the docs do live here: https://console.stage.redhat.com/beta/docs/api/content-sources

Hyperkid123 (approved):

PR #1378, opened 2023-01-17
Add ACS to left navbar for customers

kylape:
Now that ACS is doing field trials, it should show up in the left navbar for customers, not just folks with @redhat.com email addresses.

cc @vjwilson @sachaudh @rukletsov

### DIFF
--- a/chrome/application-services-navigation.json
+++ b/chrome/application-services-navigation.json
@@ -121,12 +121,6 @@
     {
       "title": "Advanced Cluster Security",
       "expandable": true,
-      "permissions": [
-        {
-          "method": "withEmail",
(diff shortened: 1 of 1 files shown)


kylape:
please hold off on merging until one of the three cc'd folks confirm this is what we want.

vjwilson (changes requested):
It's my understanding that we need to limit access to this to customers who filled out a request form. (I'm not sure if there's any contractual issues--like an NDA--but I want reassurance before we take this step.)

vjwilson (approved):
Doron says release it to the world.

Hyperkid123 (approved):

PR #1386, opened 2023-01-23
chore: MGDCTRS-1776 remove beta flag (#1383) (prod-stable)

gashcrumb:
This change removes the beta flag for connectors from the navigation.

I'll comment when we're ready to merge this.


gashcrumb:
This can be merged whenever is convenient @Hyperkid123, thanks so much!

Hyperkid123 (approved):

PR #1412, opened 2023-02-13
add fakamai settings for new rhosak-ui

riprasad:


chrome/fed-modules.json
```diff
@@ -463,13 +463,13 @@
             }
         ]
     },
-    "rhosakUi": {
-        "manifestLocation": "/apps/rhosak-ui/fed-mods.json",
```
riprasad, line 467: @Hyperkid123 @karelhala is this the internal CDN path where the assets are copied from the github build-repo to akamai/fakamai? How do we know where does our assets lie in akamai/fakamai storage? which configuration decides this location?

riprasad: okay. I figured it out. It's here https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/Jenkinsfile

The APP_NAME is defined here https://gitlab.cee.redhat.com/mk-ci-cd/rhosak-ui/-/blob/mk-release/build_deploy.sh#L23

Hyperkid123 (approved):

PR #1428, opened 2023-02-28
Enable user-access for Fedramp environment

aneelac22:


Hyperkid123:
@aneelac22 the fedramp flag has to be at the top level of the module definition, not on a route. Let me know if we have to change that.

aneelac22:
> @aneelac22 the fedramp flag has to be at the top level of the module definition, not on a route. Let me know if we have to change that.

Sorry, I don't need these changes, closing it

PR #1439, opened 2023-03-17
Move full profile to module's config

karelhala:
* [x] merge after https://github.com/RedHatInsights/insights-chrome/pull/2340


Hyperkid123:
@karelhala we will also have to wait till the linked PR is in prod-stable env.

Hyperkid123:
@karelhala can we move this to the chrome service?

PR #1460, opened 2023-03-31
Add IDs to all navigations

karelhala:
### JIRA

RHCLOUD-24831


Hyperkid123 (approved):
This should also make ti to the chrome service backend.

PR #1469, opened 2023-04-11
feat(navbar): new navbar set up

adonispuente:
In accordance with these [mocks](https://www.sketch.com/s/550bd962-ca1f-42d8-8d89-d457527a059a/a/L4OJZAQ), and the exception of inventory groups, try smart managment, and learning resources (all non-complete), these are the new navbar changes

### DIFF
--- a/chrome/rhel-navigation.json
+++ b/chrome/rhel-navigation.json
@@ -1,270 +1,316 @@
 {
-    "id": "rhel",
-    "title": "Red Hat Insights",
-    "navItems": [
-      {
-        "appId": "dashboard",
-        "title": "Overview",
-        "filterable": false,
-        "href": "/insights/dashboard",
-        "product": "Red Hat Insights"
-      },
-      {
-        "groupId": "business",
-        "icon": "trend-up",
-        "title": "Red Hat Insights",
-        "navItems": [
-          {
-            "appId": "inventory",
-            "title": "Inventory",
-            "href": "/insights/inventory",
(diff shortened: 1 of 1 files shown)


adonispuente:
I was trying to take some load off platform with this but I have a question regarding steps moving forward. The checks are failing because of the new schema required from the docs. The checks want me to add href's and app Id's to nav list that are embedded in another nav list. Wont this cause users to be directed somewhere onClick even though the item is intended as a dropdown?

Hyperkid123:
@adonispuente I see you have multple nested items in your nav. This is not supported by the PF navigation components. Can you provide mocks? We might have to rope in UX.

adonispuente:
Hey, yeah. These are the links for the new nav
 https://www.sketch.com/s/550bd962-ca1f-42d8-8d89-d457527a059a/a/L4OJZAQ

karelhala:
@adonispuente this is a bit more tricky, than just changing the config. We'll have to go trough the nav changes in chrome in order to support terciary navigation.

Moreover we're moving to the new repository so we have simpler way of serving this file https://github.com/RedHatInsights/chrome-service-backend

Thank you for bringing this to our attention! We'll discuss with our PM how crutial this is so it will be added to our roadmap so we could update the chroming navigation and then bring this over to the chroming service backend.

adonispuente:
@karelhala I see, I'll go ahead and close this PR then. Hopefully it wont be too difficult to implement

